PR #2, opened 2015-10-06
Trying to not delete frames too eagerly when laying out.

Matafou:
Hi,

Here is a patch trying to avoid deleting frames too eagerly when laying out windows for PG. In particular, if one opens a second frame and then restart scripting, the second frame is not deleted if it is displaying some other buffer.

Since this touches the generic code of pg, I let this pull request in order to let people comment. I will commit this pull request in a few days if nobody objects.

### DIFF
--- a/generic/pg-assoc.el
+++ b/generic/pg-assoc.el
@@ -52,5 +52,22 @@ Dead or nil buffers are not represented in the list."
       (setq bufs (cdr bufs)))
     wins))
 
+
+(defun proof-associated-buffer-p (b) (member b (proof-associated-buffers)))
+
+
+(defun proof-filter-associated-windows (lw)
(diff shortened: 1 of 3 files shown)


cpitclaudel:
Looks good to me; I have made a few comments directly on the diff.

Matafou:
followed one of your comments, the other (comparison to nil) was wrong.

2 remarks:
1) I made answers to your remarks, but then I rebased my pull-request and the comments disappeared, maybe this is ot the way to update a pull-request...

2) There is still an eager window deletion when restarting the prover.

So maybe we can wait a bit more that I fix this too.

cpitclaudel:
Re updating pull requests: I generally add commits and squash everything in the end. I got the responses by mail though, so nothing was lost :)

2. sounds like a good idea.

PR #6, opened 2015-10-22
Prevent non-existent directories from crashing PG

cpitclaudel:
Opening a coq file in a non-existent directory (<kbd>C-x C-f /blah/test.v</kbd>) from a fresh Emacs causes PG to crash; Emacs must thn be restarted. The error is (more or less):

```
Debugger entered--Lisp error: (file-error "Setting current directory" "No such file or directory" "/blah/")
  call-process("/bin/bash" nil t nil "-c" "/usr/local/bin/coqtop -where")
  apply(call-process "/bin/bash" nil t nil ("-c" "/usr/local/bin/coqtop -where"))
  process-file("/bin/bash" nil t nil "-c" "/usr/local/bin/coqtop -where")
  shell-command-to-string("/usr/local/bin/coqtop -where")
  (condition-case err (shell-command-to-string cmd) ((debug error) (message "Error: %S" err) nil))
  (and cmd (condition-case err (shell-command-to-string cmd) ((debug error) (message "Error: %S" err) nil)))
  (let* ((prover proof-prog-name) (cmd (and proof-prog-name (concat (shell-quote-argument prover) " -where"))) (output (and cmd (condition-case err (shell-command-to-string cmd) ((debug error) (message "Error: %S" err) nil)))) (libdir (and output (replace-regexp-in-string "[
\n]+\\'" "" output)))) (if (and libdir (file-exists-p libdir)) libdir "/usr/local/lib/coq"))
  get-coq-library-directory()
  eval((get-coq-library-directory) nil)
  eval-expression((get-coq-library-directory) nil)
  funcall-interactively(eval-expression (get-coq-library-directory) nil)
  #<subr call-interactively>(eval-expression nil nil)
  ad-Advice-call-interactively(#<subr call-interactively> eval-expression nil nil)
  apply(ad-Advice-call-interactively #<subr call-interactively> (eval-expression nil nil))
  call-interactively(eval-expression nil nil)
  command-execute(eval-expression)
```

This pull request is a partial fix, but it's not perfect. I don't think the compilation code should preemptively call coqtop -where, especially not without looking at coq-prog-name. I think that this should run every time compilation is needed; then errors won't be quite as bad, and the user will be able to change coqtops without everything breaking down.

Opinions?

### DIFF
--- a/coq/coq-system.el
+++ b/coq/coq-system.el
@@ -54,10 +54,10 @@ See also `coq-prog-env' to adjust the environment."
   :group 'coq)
 
 (defun get-coq-library-directory ()
-  (let ((c (substring (shell-command-to-string (concat coq-prog-name " -where")) 0 -1 )))
-    (if (string-match "not found" c)
-        "/usr/local/lib/coq"
-      c)))
+  ;; NOTE: This code runs when the library is loaded, so failing here causes a
+  ;; file mode specification error; hence the (ignore-errors ...) wrapper
+  (or (ignore-errors (car (process-lines coq-prog-name "-where")))
+      "/usr/local/lib/coq"))
 
 (defconst coq-library-directory (get-coq-library-directory) ;; FIXME Should be refreshed more often
   "The coq library directory, as reported by \"coqtop -where\".")


Matafou:
I agree, we should check if this code is still really usefull.

2015-10-22 18:08 GMT+02:00 Clément Pit--Claudel <notifications@github.com>:

Matafou:
Clément, is this PR still useful now that you have changed the way coqtop is detected?

cpitclaudel:
It is, though it should be updated; to reproduce, try `C-x C-f ~/non-existent-directory/test.v RET`

cpitclaudel:
Updated the PR

hendriktews:
I don't think coq-compile-ignore-library-directory makes sense (because coqdep only outputs dependency information for the Coq library with -boot). Therefore I would suggest to delete coq-compile-ignore-library-directory and the use of coq-library-directory in the background compilation code.

However, coq-library-directory is also used for coq-tags. Apart from that, coq--pre-v85 runs during initialization, when the user configured coq-compile-quick to a value not compatible with 8.4. This yields a very similar error.

For fixing all these problems, what do you think about running coqtop -where and coqtop -v in the root directory if the current directory does not exist? The version and the library should not depend on where you run coqtop...

cpitclaudel:
Thanks for the suggestions; I think that would make sense.

hendriktews:
PR #133 supersedes this PR.

PR #22, opened 2015-12-31
Fix spurious scrolling of *goals* and *response* buffers

cpitclaudel:
Hi Pierre,

Could you have a quick look at this? The bug that this attempts to fix is as follows:

In a clean coq buffer,

* Type and run "SearchAbout plus."
* Scroll to the bottom of the response buffer
* M-: (proof-shell-invisible-cmd-get-result "Check nat.")

The response buffer scrolls back to the top.

See https://github.com/cpitclaudel/company-coq/issues/8 and https://github.com/cpitclaudel/company-coq/issues/32 for some background info. Maybe we should not even call the delayed output hook? But that sounds like an invasive change.

### DIFF
--- a/coq/coq.el
+++ b/coq/coq.el
@@ -2627,19 +2627,24 @@ buffer."
 First goal is displayed on the bottom of its window, maximizing the
 number of hypothesis displayed, without hiding the goal"
   (interactive)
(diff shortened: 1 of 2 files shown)


Matafou:
Nice fix. Thanks!

Issue #21, opened 2015-12-31
Incompatibility in the handling of _CoqProject files between coq_makefile and ProofGeneral


JasonGross:
As far as I can tell, it's currently impossible to design a `_CoqProject` which passes `-compat 8.4 -require Coq.Compat.Coq84` to both `coq_makefile` and ProofGeneral.  I need something along the lines of
```
-arg "-compat 8.4"
-arg "-require Coq.Compat.Coq84"
```
to satisfy `coq_makefile`, which will blindly reorder `-arg` arguments.  However, I need something along the lines of 
```
-arg -compat
-arg 8.4
-arg -require
-arg Coq.Compat.Coq84
```
to satisfy ProofGeneral, which will otherwise invoke `coqtop -emacs "-compat 8.4" "-require Coq.Compat.Coq84"`, which causes `coqtop` to barf.  I think the correct thing to do is to fix PG to do string splitting on things that come after `-arg` in `_CoqProject`s.

cpitclaudel:
At least it's... documented?

```
;; FIXME: if several options are given (within "") in a single -arg line, then
;; they are glued and passed as a single argument to coqtop. this is bad and
;; not conforming to coq_makefile. HOWEVER: this is probably a bad feature of
;; coq_makefile to allow several arguments in a single - arg "xxx yyy".
(defconst coq-prog-args-regexp
  "\\_<\\(?1:-opt\\|-byte\\)\\|-arg[[:space:]]+\\(?:\\(?1:[^\"\t\n#]+\\)\\|\"\\(?1:[^\"]+\\)\"\\)")
```

I'll look into it.

cpitclaudel:
Hey @JasonGross,

Can you try the refactor-coqproject branch? Thanks!

Matafou:
Hi,

a first look tells me that you miss the -Q option, don't you?
I will test it but it looks good to me.

P.


2016-01-06 0:34 GMT+01:00 Clément Pit--Claudel <notifications@github.com>:

JasonGross:
@cpitclaudel that branch does not build:
```
[...]

PR #26, opened 2016-01-05
Refactor the project file parsing code

cpitclaudel:
Hey Pierre,

I've taken a stab at fixing #21. Does this look good to you?
Looking at the docs of coq_makefile, it seems it never supported -Q or -R with one option, so I dropped support for that. This makes the code much simpler.

Cheers,
Clément.

### DIFF
--- a/coq/coq-par-compile.el
+++ b/coq/coq-par-compile.el
@@ -20,8 +20,9 @@
 ;; - check what happens if coq-par-coq-arguments gets a bad load path
 ;; - on error, try to put location info into the error message
 ;;
-;; Note that all argument computations inherit coq--pre-v85: when changing
-;; compilers, all compilation jobs must be terminated.
+;; Note that all argument computations inherit `coq-autodetected-version': when
+;; changing compilers, all compilation jobs must be terminated.  This is
+;; consistent with the fact that the _CoqProject file is not reparsed.
 
 (eval-when-compile
   (require 'proof-compat))
@@ -459,7 +460,7 @@ belonging to the circle."
 Argument COQ-LOAD-PATH must be `coq-load-path' from the buffer
 that triggered the compilation, in order to provide correct
 load-path options to coqdep."
-  (nconc (coq-coqdep-prog-args coq-load-path (file-name-directory lib-src-file) coq--pre-v85)
+  (nconc (coq-coqdep-prog-args coq-load-path (file-name-directory lib-src-file) (coq--pre-v85))
          (list lib-src-file)))
 
 (defun coq-par-coqc-arguments (lib-src-file coq-load-path)
@@ -467,7 +468,7 @@ load-path options to coqdep."
 Argument COQ-LOAD-PATH must be `coq-load-path' from the buffer
 that triggered the compilation, in order to provide correct
 load-path options to coqdep."
-  (nconc (coq-coqc-prog-args coq-load-path (file-name-directory lib-src-file) coq--pre-v85)
+  (nconc (coq-coqc-prog-args coq-load-path (file-name-directory lib-src-file) (coq--pre-v85))
          (list lib-src-file)))
 
 (defun coq-par-analyse-coq-dep-exit (status output command)
--- a/coq/coq-par-compile.el
+++ b/coq/coq-par-compile.el
@@ -19,7 +19,9 @@
 ;; - add option coq-par-keep-compilation-going
 ;; - check what happens if coq-par-coq-arguments gets a bad load path
 ;; - on error, try to put location info into the error message
-;; 
+;;
+;; Note that all argument computations inherit coq--pre-v85: when changing
+;; compilers, all compilation jobs must be terminated.
 
 (eval-when-compile
   (require 'proof-compat))
@@ -450,25 +452,23 @@ belonging to the circle."
     (nreverse result)))
 
 
-;;; map coq module names to files, using synchronously running coqdep 
+;;; map coq module names to files, using synchronously running coqdep
 
 (defun coq-par-coqdep-arguments (lib-src-file coq-load-path)
   "Compute the command line arguments for invoking coqdep on LIB-SRC-FILE.
 Argument COQ-LOAD-PATH must be `coq-load-path' from the buffer
 that triggered the compilation, in order to provide correct
 load-path options to coqdep."
-  (nconc ;(coq-include-options lib-src-file coq-load-path)
-   (coq-coqdep-prog-args lib-src-file coq-load-path)
-   (list lib-src-file)))
+  (nconc (coq-coqdep-prog-args coq-load-path (file-name-directory lib-src-file) coq--pre-v85)
+         (list lib-src-file)))
 
 (defun coq-par-coqc-arguments (lib-src-file coq-load-path)
   "Compute the command line arguments for invoking coqdep on LIB-SRC-FILE.
 Argument COQ-LOAD-PATH must be `coq-load-path' from the buffer
 that triggered the compilation, in order to provide correct
 load-path options to coqdep."
-  (nconc ;(coq-include-options lib-src-file coq-load-path)
-   (coq-coqc-prog-args lib-src-file coq-load-path)
-   (list lib-src-file)))
+  (nconc (coq-coqc-prog-args coq-load-path (file-name-directory lib-src-file) coq--pre-v85)
+         (list lib-src-file)))
 
 (defun coq-par-analyse-coq-dep-exit (status output command)
   "Analyse output OUTPUT of coqdep command COMMAND with exit status STATUS.
--- a/coq/coq-seq-compile.el
+++ b/coq/coq-seq-compile.el
@@ -74,7 +74,7 @@ dependencies are absolute too and the simplified treatment of
 break."
   (let ((coqdep-arguments
          ;; FIXME should this use coq-coqdep-prog-args?
-         (nconc (coq-include-options coq-load-path (file-name-directory lib-src-file) coq--pre-v85)
+         (nconc (coq-include-options coq-load-path (file-name-directory lib-src-file) (coq--pre-v85))
 		(list lib-src-file)))
         coqdep-status coqdep-output)
     (if coq-debug-auto-compilation
@@ -113,7 +113,7 @@ Display errors in buffer `coq-compile-response-buffer'."
   (message "Recompile %s" src-file)
   (let ((coqc-arguments
          (nconc
-          (coq-coqc-prog-args coq-load-path (file-name-directory src-file) coq--pre-v85)
+          (coq-coqc-prog-args coq-load-path (file-name-directory src-file) (coq--pre-v85))
 	  (list src-file)))
         coqc-status)
     (coq-init-compile-response-buffer
@@ -122,7 +122,7 @@ Display errors in buffer `coq-compile-response-buffer'."
         (message "call coqc arg list: %s" coqc-arguments))
     (setq coqc-status
           (apply 'call-process
-           coq-compiler nil coq-compile-response-buffer t coqc-arguments))
+                 coq-compiler nil coq-compile-response-buffer t coqc-arguments))
     (if coq-debug-auto-compilation
         (message "compilation %s exited with %s, output |%s|"
                  src-file coqc-status
--- a/coq/coq-seq-compile.el
+++ b/coq/coq-seq-compile.el
@@ -73,7 +73,8 @@ dependencies are absolute too and the simplified treatment of
 `coq-load-path-include-current' in `coq-include-options' won't
 break."
   (let ((coqdep-arguments
-         (nconc (coq-include-options lib-src-file coq-load-path)
+         ;; FIXME should this use coq-coqdep-prog-args?
+         (nconc (coq-include-options coq-load-path (file-name-directory lib-src-file) coq--pre-v85)
 		(list lib-src-file)))
         coqdep-status coqdep-output)
     (if coq-debug-auto-compilation
@@ -111,8 +112,8 @@ break."
 Display errors in buffer `coq-compile-response-buffer'."
   (message "Recompile %s" src-file)
   (let ((coqc-arguments
-         (nconc ;(coq-include-options src-file coq-load-path)
-	  (coq-coqc-prog-args src-file coq-load-path)
+         (nconc
+          (coq-coqc-prog-args coq-load-path (file-name-directory src-file) coq--pre-v85)
 	  (list src-file)))
         coqc-status)
     (coq-init-compile-response-buffer
--- a/coq/coq-system.el
+++ b/coq/coq-system.el
@@ -235,30 +235,32 @@ This setting is only relevant with Coq < 8.5."
   :safe 'booleanp
   :group 'coq-auto-compile)
 
-(defun coq-option-of-load-path-entry (entry &optional pre-v85)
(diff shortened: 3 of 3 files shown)


Matafou:
Be careful I just commited a refactoring myself without realizing it would interact with yours.

Except that you seem to not deal with -Q option it looks good to me.

Matafou:
Mmmh github message was misleading, no conflict there.

Matafou:
OK, sorry for not having pushed this before.

I refactored the coq directory a bit some time ago but didn't push it to PG.

coq-system.el is now an early file to load containing things shared by several other files. In particular coq_project stuff. Your own refactoring must be put in this file I guess.

Matafou:
Actually you need to put al this in coq-system.el. That is all.

And add -Q option.

Matafou:
Actually there are other problems due to the changes of -I -Q options in v8.5.

-I now only changes caml path, -Q foo "" replaces more or less -I foo, but can if the second argument is not empty then it is a recursive directory add without import.

Therefore the load-path variable is made of more complex things than just singleton and pairs. I use some symbols to categorize things. coq-load-path must be precise enough to allow the compile-require stuff to work.
--------------
See the documentation in coq-system.el (function coq-load-path):

 "Non-standard coq library load path.
This list specifies the LoadPath extension for coqdep, coqc and
coqtop. Usually, the elements of this list are strings (for
\"-I\") or lists of two strings (for \"-R\" dir path and
\"-Q\" dir path).

The possible forms of elements of this list correspond to the 4
forms of include options ('-I' '-Q' and '-R'). An element can be

  - A list of the form '(ocamlimport dir)', specifying a
    directory to be added to ocaml path ('-I').
  - A list of the form '(rec dir path)' (where dir and path are
    strings) specifying a directory to be recursively mapped to the
    logical path 'path' ('-R dir path').
  - A list of the form '(recnoimport dir path)' (where dir and
    path are strings) specifying a directory to be recursively
    mapped to the logical path 'path' ('-Q dir path'), but not
    imported (modules accessible for import with qualified names
    only).
  - A list of the form '(norec dir)', specifying a directory
    to be mapped to the empty logical path ('-Q dir \"\"').

For convenience the symbol 'rec' can be omitted and entries of
the form '(dir path)' are interpreted as '(rec dir path)'.

Under normal circumstances this list does not need to
contain the coq standard library or \".\" for the current
directory (see `coq-load-path-include-current').

WARNING: if you use coq <= 8.4, the meaning of these options is
not the same (-I is for coq path).
"
-----------------------------

cpitclaudel:
I see, thanks for the detailed response :) Was -Q ever documented with coq_makefile?

Matafou:
What documentation do you mean?

coq_makefile -h mentions it and the coq manual (8.5) mentions the fact that
coq_makefile takes the options of coqtop.

P.

2016-01-06 16:17 GMT+01:00 Clément Pit--Claudel <notifications@github.com>:

cpitclaudel:
I meant `coq_makefile -h`. I didn't mention it in 8.4, and I failed to check in 8.5; sorry!

cpitclaudel:
Hi Pierre,

Thanks for your comments. I've updated the PR, but it still needs work. Can you check that I got the logic right?

The docstring says that there should be one set of options per line, but coq_makefile doesn't require it. If we can indeed assume that, the parsing would be easier. But the current solution should be pretty robust.

Is there really no way that we can get uniform behaviour between 8.4 and 8.5? Maybe a dynamic version check? Instead of a relying on an option.

cpitclaudel:
Btw, should -I give "ocamlimport" or not? The docstring seems to suggest both.

cpitclaudel:
Btw, you can run `coq-test-project-refactoring` in a _CoqProject buffer to compare the output of my code to the existing code. I'll remove that function before merging.

Matafou:
It should in v8.5. and it should give norec in v8.4.

P.

2016-01-06 19:33 GMT+01:00 Clément Pit--Claudel <notifications@github.com>:

cpitclaudel:
I'm confused: there's no mention of norec in the source anywhere right now.

Matafou:
Salut Clément,

effectivement c'est nonrec et pas norec. C'est dans coq-system.el. Mais vu
tes autres message je crois que tu as trouvé.

P.

2016-01-08 0:02 GMT+01:00 Clément Pit--Claudel <notifications@github.com>:

cpitclaudel:
Non j'étais vraiment confus. Mais c'est compris maintenant :) Merci pour ton aide.

cpitclaudel:
I've made more progress with this. Here are some notes on the refactoring:

* Should we mark coq-load-path-include-current obsolete? It only affects 8.4 at the moment.
* I didn't change the format of coq-load-path. I think we should discourage users from editing that variable, however.
* There were strange things in `coq-options-of-loadpath-entry':
  * For 8.5, it never generated -I (it didn't handle 'ocamlimport at all). Was that normal?
  * For 8.5, the code handled 'nonrec and 'recnoimport exactly the same. I don't think we need to handle nonrec in 8.5, correct?
  * -Q "" was special-cased, but I don't really understand why.
* I don't handle recnoimport in 8.4, and I don't handle nonrec in 8.5. Is that OK?

Matafou:
2016-01-09 20:16 GMT+01:00 Clément Pit--Claudel <notifications@github.com>:
>
> I've made more progress with this. Here are some notes on the refactoring:
>
> Should we mark coq-load-path-include-current obsolete? It only affects 8.4 at the moment.

Indeed in coq-8.5 coqdep behaves constistently with coqc/coqtop: the
current directory is always in the path (it is added (with empty path)
if it is not already by some option). coq-load-path-include-current
soudl indeed be marked obsolete for coq-8.5.

> I didn't change the format of coq-load-path. I think we should discourage users from editing that variable, however.
> There were strange things in `coq-options-of-loadpath-entry':

> For 8.5, it never generated -I (it didn't handle 'ocamlimport at all). Was that normal?

No it isn't. The case ocamlimport is missing indeed.

> For 8.5, the code handled 'nonrec and 'recnoimport exactly the same. I don't think we need to handle nonrec in 8.5, correct?

Incorrect, the "auto compile Require" part due to Hendrik uses
load-path to locate files corrresponding to module names. So I am
pretty sure replacing norec by recnoimport would be bad because of the
special treatment of -Q foo "". See below.


> -Q "" was special-cased, but I don't really understand why.

You mean -Q foo "" ?

-Q foo "" is special in coq-8.5: it corresponds to the old "-I"
option: it is not recursive *at all*. Whereas -Q foo bar means
"require recursively with non recursive Import at level 1".


>
> I don't handle recnoimport in 8.4,

yes

> and I don't handle nonrec in 8.5. Is that OK?

No.

cpitclaudel:
Cool, tanks for the comments.

2016-01-09 20:16 GMT+01:00 Clément Pit--Claudel <notifications@github.com>:
>> I've made more progress with this. Here are some notes on the refactoring:
>> Should we mark coq-load-path-include-current obsolete? It only affects 8.4 at the moment.

> Indeed in coq-8.5 coqdep behaves constistently with coqc/coqtop: the
current directory is always in the path (it is added (with empty path)
if it is not already by some option). coq-load-path-include-current
soudl indeed be marked obsolete for coq-8.5.

Thanks

>> I didn't change the format of coq-load-path. I think we should discourage users from editing that variable, however.
>> There were strange things in `coq-options-of-loadpath-entry':

>> For 8.5, it never generated -I (it didn't handle 'ocamlimport at all). Was that normal?

> No it isn't. The case ocamlimport is missing indeed.

Ok

>> For 8.5, the code handled 'nonrec and 'recnoimport exactly the same. I don't think we need to handle nonrec in 8.5, correct?

> Incorrect, the "auto compile Require" part due to Hendrik uses
load-path to locate files corrresponding to module names. So I am
pretty sure replacing norec by recnoimport would be bad because of the
special treatment of -Q foo "". See below.

Got it. When do we create a nonrec entry in coq-load-path though? When the makefile says -Q foo ""? And do we then send -Q foo "" to e.g. coqtop? If so then why do we need the special treatment? 

>> -Q "" was special-cased, but I don't really understand why.

> You mean -Q foo "" ?

Yes, sorry :)

> -Q foo "" is special in coq-8.5: it corresponds to the old "-I"
> option: it is not recursive *at all*. Whereas -Q foo bar means
> "require recursively with non recursive Import at level 1".

Wait, am I wrong to think that -Q foo "" is the same as -Q foo bar with bar == ""?

>> I don't handle recnoimport in 8.4,

> yes

>> and I don't handle nonrec in 8.5. Is that OK?

> No.

The thing that confuses me is that my current parsing code never creates 'nonrec entries in coq-load-path to begin with. If it sees -Q "a" "b" it records 'recnoimport "a" "b", and if it sees -Q "a" "" it records 'recnoimport "a" "". Then when it processes coq-load-path to generate coq-args, it produces -Q "a" "b" and -Q "" respectively.

I can add a special case for 'nonrec, but is it expected to lead to different behaviour than shown above?

Matafou:
When the second argument of -Q is empty there is a special case yes. The
directory is put in the load path (with empty logical path) but not its
subdirs. If the second arg is non empty then subdirs are added to the load
path too (with load path prefixed by the second argument).

-Q foo "" is discouraged but it helps by being equivalent to the ex -I
option.

P


Le lundi 11 janvier 2016, Clément Pit--Claudel <notifications@github.com> a
écrit :

Matafou:
Remember that we use coq-load-path also in coq-compile and I believe (will
check) that it needs a precise description of where to find modules.
Independently of coqtop/coqdep.

P

Le lundi 11 janvier 2016, Pierre Courtieu <pierre.courtieu@gmail.com> a
écrit :

> When the second argument of -Q is empty there is a special case yes. The
> directory is put in the load path (with empty logical path) but not its
> subdirs. If the second arg is non empty then subdirs are added to the load
> path too (with load path prefixed by the second argument).
>
> -Q foo "" is discouraged but it helps by being equivalent to the ex -I
> option.
>
> P
>
>
> Le lundi 11 janvier 2016, Clément Pit--Claudel <notifications@github.com

Matafou:
After checking coq-compile I think you are right it should be harmless
to remove nonrec in the current state of the ocde. Even if I think it
is a bit dangerous for the future (if we want to use load-path for
something else we will need to remember this strange case of -Q foo
"").

So finally:

* -I foo
** 8.4  nonrec foo or recnoimport  foo "" , see below
** 8.5 ocmalimport

* -R foo bar
** 8.4: rec foo bar
** 8.5: rec foo bar

* -R foo -as bar
** 8.4: rec foo bar
** 8.5: obsolete

* -Q foo ""
** 8.4: N/A
** 8.5: nonrec foo or recnoimport foo ""
  recnoimport is somehow wrong here: there is not "rec" in this case.
  BUT: since even coq-compile code does not seem to use this info
except by calling coqc and coqdep on it, it should be safe to use
recnoimport.

* -Q foo bar
** 8.4: N/A
** 8.5: recnoimport foo bar

* -I foo -as bar
** 8.4: recnoimport (not exact but very few use this anyway)
** 8.5: obsolete


* -arg
** we should be compatible with coq_makefile here.

P.













2016-01-11 16:09 GMT+01:00 Pierre Courtieu <pierre.courtieu@gmail.com>:

cpitclaudel:
Ah, thanks, that's very clear. There's just one thing:

At the moment 'nonrec is documented to take two arguments. Do we want to change this? You examples suggest that it should now take only one argument, which currently is the behaviour assigned to unannotated strings, not 'nonrec.

Matafou:
In 8.4 we need 2 arguments (-I foo -as bar), the second being optional (And
the current master is bugged isn't it?).
In 8.5 we don't need the second argument (if we keep nonrec at all).
P.

2016-01-11 18:09 GMT+01:00 Clément Pit--Claudel <notifications@github.com>:

cpitclaudel:
> In 8.4 we need 2 arguments (-I foo -as bar), the second being optional (And
> the current master is bugged isn't it?).

Current master assumes that nonrec always takes two arguments. It uses 'ocamlimport for the single argument case.

I think it's fine to not use nonrec with 8.5. This leads to:

```
* -I foo
** 8.4 ocamlimport foo
** 8.5 ocmalimport foo

* -R foo bar
** 8.4: rec foo bar
** 8.5: rec foo bar

* -R foo -as bar
Not supported when parsing a CoqProject file; I don't think coq_makefile 
ever had this documented, and master currently doesn't support it.

* -Q foo ""
** 8.4: N/A
** 8.5: recnoimport foo "" or just "foo"

* -Q foo bar
** 8.4: N/A
** 8.5: recnoimport foo bar

* -I foo -as bar
** 8.4: nonrec foo bar (never parsed from a CoqProject though)
** 8.5: obsolete
```

Matafou:
OK with me.

Just put in a comment that:
1) ocamlimport is not about ocaml when in 8.4
2) -Q "" is actually not recursive.
So that future people do not mess with this tags.

P.


2016-01-11 18:40 GMT+01:00 Clément Pit--Claudel <notifications@github.com>:

Matafou:
For the record, coq-compile modules compute dependencies without modelling the coq file loading logic, which is a good abstraction thing even if it overuses coqdep.

To compute the physical path of a logical module foo.bar, it calls coqdep on single lined temporary file containing Require foo.bar and then parse the dependencies generated.
`
(defun coq-seq-map-module-id-to-obj-file (module-id span)
  "Map MODULE-ID to the appropriate coq object file.
The mapping depends of course on coq-load-path. The current
implementation invokes coqdep with a one-line require command.
This is probably slower but much simpler than modelling coq file
loading inside ProofGeneral.
...`

cpitclaudel:
I've updated the comment; thanks for checking the rest of the code! I believe we should merge this before 8.5 on Friday ^^

cpitclaudel:
Oups, there's an issue in that last commit

cpitclaudel:
Fixed :)

JasonGross:
This seems to work well for me; thanks!

JasonGross:
Sorry, actually, this gives me
```
In toplevel form:
coq/coq-compile-common.el:17:1:Error: Junk in args (:type (repeat (choice (string :tag "simple directory without path (-Q \"\") in 8.5, -I in 8.4") (list :tag "recursive directory with path (-R ... ...)" (const rec) (string :tag "directory") (string :tag "log path"))
(list :tag "recursive directory without recursive import with path (-Q ... ...)" (const recnoimport) (string :tag "directory") (string
 :tag "log path")) (list :tag "compatibility for of -I (-I ... -as ... in coq<=8.4)" (const nonrec) (string :tag "directory") (string
:tag "log path")) (list :tag "ocaml path (-I ...)" (const ocamlimport) (string :tag "directory") (string :tag "log path")))) :safe coq
-load-path-safep :group coq-auto-compile)
make[1]: *** [coq/coq-compile-common.elc] Error 1
```
with GNU Emacs 24.3.1

cpitclaudel:
Indeed, because I fixed the issue but forgot to push :/

JasonGross:
Works now, thanks!

cpitclaudel:
Great

cpitclaudel:
Hey Pierre,

Is there anything blocking a merge at this point, or should I go ahead?

Clément.

Matafou:
Nothing blocking AFAIK. Go ahead.
P.

2016-01-14 17:08 GMT+01:00 Clément Pit--Claudel <notifications@github.com>:

cpitclaudel:
Done! Thanks for testing :)

PR #27, opened 2016-01-07
Add a .travis.yml file for continuous integration emacs 23, 24

JasonGross:
This is based on (cargo culted from)
https://github.com/rolandwalker/emacs-travis/blob/master/.travis.yml.  Someone
will have to go to https://travis-ci.org/profile/ProofGeneral to enable Travis
for PG.  Do we want to require emacs-snapshot to succeed, or allow it to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,164 @@
+###
+### Notes
+###
+### Adapted from https://github.com/rolandwalker/emacs-travis/blob/master/.travis.yml
+###
+### The travis web interface may choke silently and fail to
+### update when there are issues with the .travis.yml file.
+###
+### The "travis-lint" command-line tool does not catch all
+### errors which may lead to silent failure.
+###
+### Shell-style comments in this file must have "#" as the
+### *first* character of the line.
+###
+
+###
+### language
+###
+
+# travis-lint no longer permits this value
+# language: emacs-lisp
+
+###
+### defining the build matrix
+###
+### ===>                                                       <===
+### ===> each variation in env/matrix will be built and tested <===
+### ===>                                                       <===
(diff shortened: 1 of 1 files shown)


Matafou:
Nice!

Enabled.


2016-01-08 0:51 GMT+01:00 Jason Gross <notifications@github.com>:

cpitclaudel:
Thanks Jason. You can remove the stuff about Emacs 22 and 23; I don't think we're interested in supporting them. Debian stable ships Emacs 24, so that ought to be fine :)

OOTH, I don't think we want to allow failures on emacs-snapshot, and it would be nice to test with emacs 24.3, 24.4, and 24.5. I don't know if that's possible using Ubuntu's package manager. I usually do it by recompiling emacs (see https://github.com/FStarLang/fstar-mode.el/blob/master/.travis.yml for example).

Recompiling Emacs also has the advantage that we can run on Travis' container-based architecture, where we don't have sudo.

JasonGross:
Do you want to just make a fresh pull request taking the relevant stuff from fstar-mode.el?  It looks cleaner, and I didn't actually put much work into this.

cpitclaudel:
I don't have much time at the moment, so I won't be able to look at this before after POPL and the CAV deadlines at least :/

Matafou:
Jason, we will add GPL licence and give copyright to fsf. Do you have objections?
P.

cpitclaudel:
IIUC, this is minor adaptations over code that I wrote for F*-mode; this qualifies for the FSF exemption, since I already have copyright papers on file.

JasonGross:
I have no objections in any case

PR #50, opened 2016-02-13
Implement support for "From ... Require ..."

### LINKED ISSUE #49: Automatic compliation of dependencies does not work with Coq 8.5's "From ... Require Import..."
When a file starts, for example, with
```
From program_logic Require Import ownership.
```
ProofGeneral fails to notice that it has to re-compile the file, and then immediately fails because of inconsistent assumptions about other modules.

RalfJung:
This is my first ever lisp program, so I probably violated every style guideline in existence - I tried to match the surroundings though.

Also, the same still needs to be sine for sequential compilation mode. But this fix means I can go on doing actual work, so that's what I will do next ;-) . If you only want to merge this once sequential mode is also implemented - I can do that, when I feel bored and want to do something different, but not right now.

Fixes #49

### DIFF
--- a/coq/coq-compile-common.el
+++ b/coq/coq-compile-common.el
@@ -301,11 +301,12 @@ the latter condition into an error, then set this variable to
   :group 'coq-auto-compile)
 
 (defconst coq-require-command-regexp
-  "^Require[ \t\n]+\\(Import\\|Export\\)?[ \t\n]*"
+  "^\\(From[ \t\n]+\\([A-Za-z0-9_']+\\)[ \t\n]+\\)?Require[ \t\n]+\\(Import\\|Export\\)?[ \t\n]*"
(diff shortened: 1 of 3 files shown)


RalfJung:
I just couldn't leave this like that... so I pushed the changes to also compute the module ID properly in sequential mode. ;-)

Matafou:
Thanks a lot Ralf!

Will merge as soon as I have time for it.

P.


2016-02-13 15:56 GMT+01:00 Ralf Jung <notifications@github.com>:

hendriktews:
Thanks for the bug report and the fix. I would like to have a look at the problem and then merge this pull request.

RalfJung:
I updated the PR to fix a problem in sequential mode.

hendriktews:
Thanks for the contribution, the changes somehow made it into master without merging this PR.

PR #79, opened 2016-06-18
coq-load-path docs: norec -> nonrec

tbrk:
Otherwise coq-load-path-safep fails. It seems to be nonrec everywhere else.

### DIFF
--- a/doc/ProofGeneral.texi
+++ b/doc/ProofGeneral.texi
@@ -4718,7 +4718,7 @@ forms of include options ('-I' and @code{'-R'}). An element can be
   - A list of the form '(rec dir path)' (where dir and path are
     strings) specifying a directory to be recursively mapped to the
     logical path @code{'path'} ('-R dir -as path').
-  - A list of the form '(norec dir path)', specifying a directory
+  - A list of the form '(nonrec dir path)', specifying a directory
     to be mapped to the logical path @code{'path'} ('-I dir -as path').
 @end lisp
 For convenience the symbol @code{'rec'} can be omitted and entries of


cpitclaudel:
Thanks, merged.
Btw, is there a reason why you're setting this by hand instead of using a _CoqProject?

tbrk:
I'm working with a development that builds on both AbsInt/CompCert and PrincetonUniversity/VST.
CompCert provides a script that manually generates coq-prog-args, while the VST provides either a _CoqProject or a .loadpath with command-line arguments. I was playing with the idea of adapting the CompCert script to generate a command-line for both projects and my own, but I'm still having trouble getting this to work properly with compile-before-require. Maybe I should give up and just generate a _CoqProject for CompCert and write a script to merge multiple _CoqProject files?

In short, is there a way to include _CoqProject files from subdirectories into a master _CoqProject file?

cpitclaudel:
Thanks for the clarification. I don't think there is a way to merge; I'm not sure what the semantics would/should be.

Matafou:
For the record: as far as I tested Compcert can be compiled with a very simple _CoqProject + coq_makefile.
I am quite sure you can have a _CoqProject that works for both projects, which is probably the cheapest way to make compile-before-require operational.

Here the _CoqProject file I used for compcert (this is from Compcert directory.

-R ia32 compcert.ia32
-R lib compcert.lib
-R common compcert.common
-R backend compcert.backend
-R cfrontend compcert.cfrontend
-R driver compcert.driver
-R flocq compcert.flocq
-R exportclight compcert.exportclight
-R cparser compcert.cparser

if you want to have your _CoqPoject file one level below (to have VST and Compcert side by side), just replace

-R foo bar
by
-R Compcert/foo bar:

P.

tbrk:
Thank you. Yes, I also tried a _CoqProject file. I can confirm that the following works:
```
-R CompCert/lib compcert.lib
-R CompCert/common compcert.common
-R CompCert/ia32 compcert.ia32
-R CompCert/backend compcert.backend
-R CompCert/cfrontend compcert.cfrontend
-R CompCert/driver compcert.driver
-R CompCert/flocq compcert.flocq
-R CompCert/exportclight compcert.exportclight
-R CompCert/cparser compcert.cparser

-R VST/msl msl
-R VST/sepcomp sepcomp
-R VST/veric veric
-R VST/floyd floyd
```
I still thought it was worth updating the emacs docs since I lost a bit of time trying to understand why `norec` wouldn't work.

cpitclaudel:
> I still thought it was worth updating the emacs docs since I lost a bit of time trying to understand why norec wouldn't work.

Absolutely! Thanks very much :)

cpitclaudel:
(And sorry for the wasted time)

tbrk:
Not at all! Thanks to both of you (and all the others) for maintaining/developing this useful tool.

PR #85, opened 2016-07-03
Highlight Existing Class like Existing Instance

### LINKED ISSUE #81: Existing Class is not completely colorized
At least not the way that Existing Instance is.

JasonGross:
Closes #81


cpitclaudel:
Thanks!

PR #87, opened 2016-07-04
Fix (next-undo-elt) to return a relevant undo element w.r.t (undo-delta)

### LINKED ISSUE #39: Undo does not revert the processed region
If I `C-/`, and this deletes a word from the already processed region, the region is not reverted.

erikmd:
This a patch to fix #39

I have tested my branch only with GNU Emacs 24.3.1, so more testing is welcome.
Ref: https://github.com/ProofGeneral/PG/issues/39#issuecomment-230180627
Cc: @JasonGross @Matafou @cpitclaudel


Matafou:
Checked. It fixes the bug in emacs-24.
Thank Erik.
P.

PR #98, opened 2016-08-03
Ensure PG overlays have pg-span property

### LINKED ISSUE #97: Retracting to a point with multiple overlays fails
I can reproduce this in Spacemacs by attempting to make any change starting at an open parenthesis, which fails with "wrong type argument: number-or-marker-p, nil", and nothing is backtracked. (This is due to smartparens in particular, nothing Spacemacs-specific.)

I tracked down the error to coq-find-and-forget, https://github.com/ProofGeneral/PG/blob/master/coq/coq.el#L729, where the check `(= proofdepth 0)` fails since `proofdepth` is nil.

Furthermore, this occurs because the wrong overlay is being passed to this function (the span argument). This is due to proof-retract-until-point using `(span-at (point) 'type)` to get the span, when it should really search `(spans-at-point-prop (point) 'type)` for one that is a ProofGeneral span. For reference, here's what `(mapcar 'overlay-properties (spans-at-point-prop (point) 'type))` evaluates to on a paren for me (with some spacing adding for readability):

```
((type show-pair priority 1000 face sp-show-pair-match-face)

(proofnum 0 proofstack nil statenum 51 pg-helpspan #<overlay from 1 to 25 in test.v>
isearch-open-invisible-temporary pg-open-invisible-span isearch-open-invisible pg-open-invisible-span
span-delete-actions ((lambda nil (pg-remove-element (quote command) (quote command-9))))
rawname nil name "command-9" id command-9 idiom command
cmd "Definition foo := (1+1)." type vanilla))
```

The first span belongs to smartparens, while the second is what PG is looking for. I'm not sure what the best overlay property to filter for is among all these.

tchajed:
Fixes #97 and (probably) #8. Thanks @cpitclaudel!

### DIFF
--- a/generic/pg-response.el
+++ b/generic/pg-response.el
@@ -393,23 +393,23 @@ Returns non-nil if response buffer was cleared."
    (t
     (let (start end)
       (with-current-buffer proof-response-buffer
-	(setq buffer-read-only nil)
-	;; da: I've moved newline before the string itself, to match
-	;; the other cases when messages are inserted and to cope
-	;; with warnings after delayed output (non newline terminated).
-	(goto-char (point-max))
-	;; insert a newline before the new message unless the
-	;; buffer is empty or proof-script-insert-newlines is nil
-	(unless (or (not proof-script-insert-newlines)
-		    (eq (point-min) (point-max)))
-	  (newline))
-	(setq start (point))
-	(insert str)
-	(unless (bolp) (newline))
-	(when face
-	  (overlay-put
-	   (make-overlay start (point-max))
-	   'face face))
+        (setq buffer-read-only nil)
+        ;; da: I've moved newline before the string itself, to match
+        ;; the other cases when messages are inserted and to cope
+        ;; with warnings after delayed output (non newline terminated).
+        (goto-char (point-max))
+        ;; insert a newline before the new message unless the
+        ;; buffer is empty or proof-script-insert-newlines is nil
+        (unless (or (not proof-script-insert-newlines)
+                    (eq (point-min) (point-max)))
+          (newline))
+        (setq start (point))
+        (insert str)
+        (unless (bolp) (newline))
+        (when face
+          (overlay-put
+           (span-make start (point-max))
+           'face face))
 
 	(setq buffer-read-only t)
 	(set-buffer-modified-p nil))))))
--- a/lib/span.el
+++ b/lib/span.el
@@ -12,8 +12,7 @@
 ;; Spans are our abstraction of extents/overlays.  Nowadays
 ;; we implement them directly with overlays.
 ;;
-;; In future this module should be used to implement the abstraction
-;; for script buffers (only) more directly.
+;; FIXME: eliminate aliases and directly use overlays
 ;;
 
 ;;; Code:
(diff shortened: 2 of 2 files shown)


cpitclaudel:
@Matafou @erikmd Can you have a look? Tej and I looked at this together today.

The new invariant is that all PG overlays have a `pg-span` property. Tej also refactored `spans-filter` as a separate function, and I pushed to get rid of these `defsubsts`, because `defsubsts` are a pain compilation-wise.

erikmd:
Hi @tchajed and @cpitclaudel,
Thanks for submitting this; I don't have a Spacemacs installed to hand so I couldn't reproduce #97; but I closely looked at the code and did several Coq proofs with the PR branch, and it LGTM.

Still, I am wondering if the few occurrences of `make-overlay` in the other PG files should be replaced with `span-make` as well?
For the record, there are mainly 5 such occurrences:

1. [generic/pg-response.el#L411](https://github.com/ProofGeneral/PG/blob/dbfe0a4f3b7be8f18ff89f5a544cb9d9ef891b70/generic/pg-response.el#L411)
2. [mmm/mmm-region.el#L435](https://github.com/ProofGeneral/PG/blob/dbfe0a4f3b7be8f18ff89f5a544cb9d9ef891b70/contrib/mmm/mmm-region.el#L435)
3. [lib/holes.el#L45](https://github.com/ProofGeneral/PG/blob/dbfe0a4f3b7be8f18ff89f5a544cb9d9ef891b70/lib/holes.el#L45)
4. [lib/unicode-tokens.el#L934](https://github.com/ProofGeneral/PG/blob/dbfe0a4f3b7be8f18ff89f5a544cb9d9ef891b70/lib/unicode-tokens.el#L934)
5. [lib/unicode-tokens.el#L964](https://github.com/ProofGeneral/PG/blob/dbfe0a4f3b7be8f18ff89f5a544cb9d9ef891b70/lib/unicode-tokens.el#L964)

(while it appears that the occurrences 1. 4. 5. won't deal with the main proof buffer).

@cpitclaudel @Matafou, what's your opinion on this?

tchajed:
Of those examples, I think only `lib/holes.el` intended to use spans, given that it uses `span-detach` and `span-detached-p`. It probably doesn't matter much in that particular case because the holes library doesn't do span filtering, where the distinction between spans and overlays becomes meaningful.

If you think that `lib/holes.el` should be updated I'm happy to update it. I'll also change `delete-overlay` to the alias `span-detach` while I'm at it.

cpitclaudel:
Thanks for looking at the patch :)

1. is safe, but it wouldn't hurt to update that use either
2. is vendored code; I wouldn't touch it
3. is @matafou's :) It seems safe, since it's detached
4, 5. are mostly dead code, but they are safe.

I vote for updating 1 and letting @matafou do TRT for 3 :)
I wouldn't add span-detach calls where they don't currently exist; the indirection makes everything harder to read.

Matafou:
We should get rid of holes anyway and use a standard skeleton stuff
instead. Or better: simply let company-Coq do this job :-).
I think nobody uses holes except me anyway.
P

Le dimanche 14 août 2016, Clément Pit--Claudel <notifications@github.com> a
écrit :

tchajed:
I updated the use of `make-overlay` in [generic/pg-response.el](https://github.com/ProofGeneral/PG/pull/98/commits/2b49e4b94e59cbb82073cfbe4fef40a811663d38#diff-9d5bd18400d922d95d801ab7d6628084R411) and left the others alone. The diff is larger than I'd like because I took the liberty of re-indenting the surrounding function.

Is this pull request ready to merge then?

cpitclaudel:
Yup, thanks :) Can you start working on the copyright assignment? This is very close to the "small change" threshold.

PR #106, opened 2016-09-01
Add {Local ,}Axioms like {Local ,}Axiom, fix Parameters

JasonGross:
I couldn't figure out whether or not to put a final `.`; I left it off as in `Axiom`

### DIFF
--- a/coq/coq-syntax.el
+++ b/coq/coq-syntax.el
@@ -411,7 +411,9 @@ so for the following reasons:
 (defvar coq-decl-db
   '(
     ("Local Axiom" nil "Local Axiom # : #" t "Local\\s-+Axiom")
+    ("Local Axioms" nil "Local Axioms # , #: #" t "Local\\s-+Axioms")
     ("Axiom" "ax" "Axiom # : #" t "Axiom")
+    ("Axioms" "axs" "Axioms # , #: #" t "Axioms")
     ("Global Variable" "gv" "Global Variable #: #." t "Global\\s-+Variable")
     ("Global Variables" "gvs" "Global Variables # , #: #." t "Global\\s-+Variables")
     ("Hint Constructors" "hc" "Hint Constructors # : #." t "Hint\\s-+Constructors")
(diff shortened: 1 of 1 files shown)


erikmd:
Hi @JasonGross, thanks for your PR.
> I couldn't figure out whether or not to put a final .

Good point! Indeed it appears that in these `coq-...-db` variables of `coq-syntax.el`, the presence (or absence) of "final dots" is definitely not uniform!

Ideally I believe that the caddr of all sublists of `coq-decl-db`, `coq-defn-db`, `coq-goal-starters-db`, `coq-queries-commands-db`, `coq-other-commands-db`, `coq-ssreflect-commands-db` should end with a dot if the corresponding string is a complete command (possibly containing `#`).
So for example: "Axiom # : #" -> "Axiom # : #." (whereas "Check" should not be replaced with "Check." but be kept as is... or replaced with "Check #.")

Of course this issue is not critical, but I see two solutions here:
1. We update all these caddr strings just now.
2. We postpone this change and wait that Paul's branch is merged, and for the moment we merge this PR after @JasonGross has updated the "final dots" for the lines already related to this PR.

@cpitclaudel and @Matafou, what is your opinion?

cpitclaudel:
I'd be in favor of of merging this now.  This should create only minimal conflicts.

Matafou:
I see no reason why Paul would have touched this file anyway, do you?
P.

2017-01-17 15:32 GMT+01:00 Clément Pit--Claudel <notifications@github.com>:

erikmd:
Hi @cpitclaudel and @Matafou,
> I'd be in favor of of merging this now.

OK.

> I see no reason why Paul would have touched this file anyway, do you?

I've just checked the history of `coq-syntax.el` in Paul's fork (using gitk) and indeed, the latest version of this file is the same in both ProofGeneral/PG and psteckler/ProofGeneral repos.

So I'm going to merge this PR (which has been open for a long time, sorry @JasonGross for that) and do another commit in master to fix all final-dots updates I was talking about in my comment https://github.com/ProofGeneral/PG/pull/106#issuecomment-273145975
All the best

Matafou:
Thanks @erikmd @JasonGross .

cpitclaudel:
Super, thanks everyone!

PR #112, opened 2016-09-18
Make shell and perl scripts executable

zmwangx:
The scripts I'm `chmod +x`'ing all have shebangs, so they should have the executable bit set. Without that, none of the `BIN_SCRIPTS` installed by `make install` into `bin/` is executable, making them pointless residents of `bin/` (unless one invokes sh or perl with the full path to the script).


erikmd:
Hi @zmwangx, thanks for reporting this.
@Matafou and @cpitclaudel, should we make a 4.4.1 release for that?

erikmd:
BTW @zmwangx, can you confirm this was the only issue with "make install" on your system, and that PG works smoothly afterwards?
(We did not double-check if the Makefile was OK with all configurations, given that the current default installation procedure https://github.com/ProofGeneral/PG/blob/master/README.md#setup does not rely on the Makefile...)
Best regards.

zmwangx:
@erikmd In fact, I noticed that if I run `make` with one thread only, the resulting scripts will be executable; only when I run parallelized `make` (`make -j4` for instance) will this bin scripts not executable issue show up, and in that case some files like `bin/proofgeneral` are actually empty.

So this PR is not a valid fix (but maybe they scripts should be executable anyway in the repository?); what should be fixed is the problem with parallelized make, or as a short term solution, you can ask users to run `make -j1`.

zmwangx:
Sorry for the noise by the way.

cpitclaudel:
We should indeed investigate parallel makes; in the meantime, I think a small note next to installation instructions would do :) Thanks for the analysis @zmwangx !

erikmd:
Hi @cpitclaudel, I've just pushed a commit that fixes the executable-bit issue.
Regarding the Makefile-related issue we'll indeed have to figure out what is going on exactly. But in the meantime, maybe it is unnecessary to add a `-j1` disclaimer in the installation instructions, as we could use the [.NOTPARALLEL](https://www.gnu.org/software/make/manual/html_node/Special-Targets.html#index-_002eNOTPARALLEL) special target of make ?
Best, EMD

cpitclaudel:
>  maybe it is unnecessary to add a -j1 disclaimer in the installation instructions, as we could use the .NOTPARALLEL special target of make ?

Yes please :) Sounds great.

PR #115, opened 2016-10-14
Refer to Apple's operating system as macOS

tchajed:
Starting with the latest version (10.12 "Sierra"), the operating system
is called macOS rather than Mac OS X.

Mac is technically the name of the hardware, not the operating system.


erikmd:
Thanks!

Issue #92, opened 2016-07-15
Compile before require from current directory failing with 8.5


YaZko:
Dear all,

I (finally) switched to Coq 8.5, and am having troubles with PG's compile before require feature. A simple Require Import Test where Test.v is in the current directory fails with the error displayed below. Note that the very same Import does not fail with 8.4 and that I am up to date with the trunk.

As I understand it, -I option has changed quite a bit between 8.4 and 8.5. However, following the documentation, this simple case should not be affected I believe.
* My coq-load-path is set to nil
* No -I options in my coq-prog-name nor coq-prog-args
* My coq-load-path-include-current is set to t (but is anyway obsolete with 8.5 if I understood the documentation correctly)

Am I missing something else, is this a bug or is a _CoqProject now always mandatory to use the CBR feature?

Thanks in advance,
Best,
Yannick

```
-*- mode: compilation; -*-

echo "Require Test." > /var/folders/xq/my2cmz3n26bf9sw4ryqk_cnh0000gn/T/ProofGeneral-coq64463VbT.v /Users/Rainette/.nix-profile/bin/coqdep -Q /Users/Rainette/Work/zakowski/trunk/Thoughts_and_Notes/Simulations/ "" /var/folders/xq/my2cmz3n26bf9sw4ryqk_cnh0000gn/T/ProofGeneral-coq64463VbT.v
*** Warning: in file /var/folders/xq/my2cmz3n26bf9sw4ryqk_cnh0000gn/T/ProofGeneral-coq64463VbT.v, library Test is required and has not been found in the loadpath!
*** Warning: in file /var/folders/xq/my2cmz3n26bf9sw4ryqk_cnh0000gn/T/ProofGeneral-coq64463VbT.v, library Test is required and has not been found in the loadpath!
/var/folders/xq/my2cmz3n26bf9sw4ryqk_cnh0000gn/T/ProofGeneral-coq64463VbT.vo /var/folders/xq/my2cmz3n26bf9sw4ryqk_cnh0000gn/T/ProofGeneral-coq64463VbT.glob /var/folders/xq/my2cmz3n26bf9sw4ryqk_cnh0000gn/T/ProofGeneral-coq64463VbT.v.beautified: /var/folders/xq/my2cmz3n26bf9sw4ryqk_cnh0000gn/T/ProofGeneral-coq64463VbT.v
[...]

PR #118, opened 2016-10-26
Several fixes

### LINKED ISSUE #70: Coq trunk + compile before require => « Invalid version syntax: 'trunk' »
When I use a trunk version of Coq together with the `compile-before-require` option turned on, then if I try to `Require` a module (at least one from the standard library), I get the following error message in the `*Messages*` buffer (and a shortened version in the mini-buffer):

> byte-code: Invalid version syntax: 'trunk'

Minimal reproduction case is the following file:
```
Require Logic.
```

hendriktews:
I had quite a long break from PG hacking. Therefore I open this PR instead of pushing directly. If nobody comments, I will merge in a few days.

The commits here are quite independent:
- fix parallel background compilation (it did not work at all for me)
- delete and gitignore TAGS
- rename doc/.cvsignore into doc/.gitignore
- fix #70


cpitclaudel:
Thanks, this looks great.  I wonder if we're starting to do to much regarding version detection, though.  For now, we just use that code to detect Coq < 8.5 vs Coq >= 8.5, right?  If so, I think we could just issue a warning when we see "trunk" as a version number, and change it to "8.5". That warning could suggest to set coq-pinned-version, too. What do you think?

hendriktews:
forced push a tiny change: commit 24858fa will fix #92

cpitclaudel:
Hendrik, did you see my comment above before merging? I think this version detection code is getting much more complex than it needs to be; considering "trunk" as 8.5 and issuing a warning when we see it would be sufficient, I think.

Also: the code your just merged checks for "trunkk" instead of "trunk". Does it actually  work?

hendriktews:
Clément Pit--Claudel <notifications@github.com> writes:

> issue a warning when we see "trunk" as a version number, and change it
> to "8.5". That warning could suggest to set coq-pinned-version, too.
> What do you think?

This is a little change, I would not argue about it. However, I
would not be motivated to do it myself, because it is clear now
already, that this piece would need continuous maintenance in the
future.

Hendrik

hendriktews:
Clément Pit--Claudel <notifications@github.com> writes:

> Hendrik, did you see my comment above before merging? I think this

Yes, and I answered and now I wonder why my answer does not show
up.

The problem that I have with changing trunk into 8.5 is that such
a piece of code will silently break in the future. Further, it
will not catch the errors from version-to-list for other strange
version strings. 

But I don't have a strong opinion on this. I thought it is good
to close this issue, but if you believe this commit is the wrong
way, then please revert it.

> Also: the code your just merged checks for "trunkk" instead of

Ah.. The best idea I had to test the else branch was to change
the condition and I forgot to undo that. The code does work, you
just don't see the trunk specific message on trunk. I will fix
this soon.

Hendrik

cpitclaudel:
>  The problem that I have with changing trunk into 8.5 is that such a piece of code will silently break in the future. Further, it will not catch the errors from version-to-list for other strange version strings.

Not really: it'll show a warning, and hopefully we can convince the Coq devs to fix the underlying problem.

> Further, it will not catch the errors from version-to-list for other strange version strings.

Do we want to? I think we'd rather hear about new weird version strings, right?

PR #127, opened 2016-12-05
A first pass at converting from CVS to git.

zhenya1007:
Just a quick pass over the Makefile, replacing usage of CVS commands
with (what I believe are) their approximate git equivalents.

My end goal is to have "make pkg" produce a fully-usable Emacs package.  It isn't quite there yet: after installing `ProofGeneral-20.tar` with `package-install-file`, doing `M-x locate-library proof-site` complains that it cannot find `proof-site`, so something needs to be fixed (also byte-compilation that runs at the package installation time generates a lot of errors and warnings).  Loading `proof-site.el` "by hand" does appear to "make things work."

Regardless of the issues with the package, I hope you can accept the conversion from CVS to git in Makefile.devel.


Makefile.devel
```diff
@@ -309,17 +293,17 @@ distcompile: FORCE
...
-##	   set version stamp in proof-site.el and ProofGeneral.spec 
+## tag:    tag the sources of working directory with GIT_RELEASENAME,
+##	   set version stamp in proof-site.el and ProofGeneral.spec
 ##         to VERSION, and edit $(DOWNLOADHTMLS)
 ##	   if VERSION matches PRERELEASE_TAG.
 ##
-tag:		
+tag:
 	@echo "*************************************************"
 	@echo " Tagging sources... (you must rerun if CVS source dirty)"
```
cpitclaudel: Leftover mention of CVS here :)

cpitclaudel:
Thanks, this looks great! Having a package-able PG would be great.
I'll let @erikmd comment, since he's better versed than I am in PG-packaging-related matters.

One issue that you'll run into is that PG has subdirectories containing autoload comments; these are not picked up by package.el, so we need to be a bit smarter and put some autoloaded initialization code in a file at the root.

erikmd:
Hi @cpitclaudel ! Thanks @zhenya1007 for submitting this. Indeed the PG Makefiles deserve some cleanup as there is no CVS around anymore. I won't have much time before the end of the week because of a deadline, but I'll try my best to review this asap.
Best regards.

zhenya1007:
Hi, @erikmd: no worries.  I am sure we will all survive if this patch doesn't get reviewed until some time next week. :-)  Thank you @cpitclaudel for the tip about autoloads trickiness: I have noticed that the ProofGeneral-autoloads.el as generated by package.el ends up being empty, and now I understand why.

cpitclaudel:
@zhenya1007 On that topic, I think you'll find the discussion and PR at https://github.com/realgud/realgud/issues/84 interesting.

cpitclaudel:
Thanks again for working on this!

zhenya1007:
Ah, thank you that's a very nice approach.  But I am not sure I understand why you want to generate the autoloads file, and then ship it.  It seems to me, that, since the `package-install-*` function(s) will call `update-directory-autoloads` in the top-level directory of the package exactly once (at package installation time), you could have the code that generates the autoloads file in a "bootstrap" file at the top directory of your package.  I will start a separate pull request to show exactly what  I mean.   Of course, it's entirely possible I am missing something.

cpitclaudel:
I don't think you're missing anything :) It just seemed safer to pre-process the autoloads than to build them dynamically, but as long as we test the process appropriately install-time generation should work too.

I'll have time to review the other PR this week-end, I think.

cpitclaudel:
Thanks again!

Makefile.devel
```diff
@@ -309,17 +293,17 @@ distcompile: FORCE
...
 	@echo "*************************************************"
 # Update the sources, this is almost always what we want to do.
-	if [ -z "$(NOCVS)" ] && [ -n "`cvs -n -q update -Pd | grep '^[MC] '`" ]; then cvs update -Pd; exit 1; fi
+	if [ -z "$(NOCVS)" ]; then git status | grep '^Your branch is up-to-date with.*master' && git status | grep 'nothing to commit, working tree clean' || exit 1 fi
```
erikmd: This line should be replaced with:
`if [ -z "$(NOCVS)" ]; then git fetch origin && git log --exit-code master...origin/master || exit 1; fi`

Makefile.devel
```diff
@@ -334,26 +318,25 @@ tag:
 		rm $$f.old; \
 	   done) \
 	fi
```
erikmd: This if-then-else (starting with `if [ $(PRERELEASE_TAG) = $(VERSION) ]; then`) should be commented-out as the PG website is now handled separately.

Makefile.devel
```diff
@@ -334,26 +318,25 @@ tag:
...
-	if [ -z "$(NOCVS)" ]; then cvs commit -m"Set version tag for new release." generic/$(VERSIONFILE) etc/ProofGeneral.spec; (cd $(HTMLDIR); cvs commit -m"Set version tag for new release." $(DOWNLOADHTMLS)); fi
-	if [ -z "$(NOCVS)" ]; then cvs tag -cF "$(CVS_RELEASENAME)"; fi
+	if [ -z "$(NOCVS)" ]; then git add generic/$(VERSIONFILE) etc/ProofGeneral.spec && git commit -m"Set version tag for new release." ; git add $(HTMLDIR)/$(DOWNLOADHTMLS) && git commit -m"Set version tag for new release."; fi
```
erikmd: You should remove the last command here, that is, remove `; git add $(HTMLDIR)/$(DOWNLOADHTMLS) && git commit -m"Set version tag for new release."`

Makefile.devel
```diff
@@ -334,26 +318,25 @@ tag:
...
+	if [ -z "$(NOCVS)" ]; then git tag -f  "$(GIT_RELEASENAME)"; fi
```
erikmd: You should replace `git tag -f  "$(GIT_RELEASENAME)"` with `git tag -a "$(GIT_RELEASENAME)"`

Makefile.devel
```diff
@@ -362,12 +345,13 @@ cvsexport:
 	rm -rf $(DISTBUILDIR)
 	mkdir -p $(DISTBUILDIR)
 	if [ -z "$(NOCVS)" ]; then \
-           (cd $(DISTBUILDIR); \
-	    cvs -d $(CVSROOT) export -kv -r "${CVS_RELEASENAME}" -d ${RELEASENAME} ${CVSNAME}) \
+	    git worktree prune; \
```
erikmd: `git worktree` is definitely overkill here.
Instead, you could replace these two lines with
`git archive --format=tar --prefix=$(RELEASENAME)/ $(GIT_RELEASENAME) | (cd "$(DISTBUILDIR)" && tar xf -)`
**and** remove the `-rm -rf "$(DISTBUILDIR)/$(RELEASENAME)/.git"` line below.

erikmd:
Hi @zhenya1007,
Another detail:
as your commit 13b859d in this PR is almost the same as your commit 9e73368 in #128, I suggest that you also add a commit that reverts 13b859d in the fork branch related to this PR (zhenya1007:master), e.g. by doing `git checkout master && git revert 13b859d`
(Once the PR will be ready for merging, I'll do a squash-and-merge so 13b859d and the revert-commit won't appear anymore.)
Thank you again!
EMD

Makefile.devel
```diff
@@ -89,14 +89,14 @@ DEVELOPERS=\
 # PRERELEASE_PREFIX is used to match PRERELEASE_TAG in sed
 # line in tag target below, which edits $(DOWNLOADHTML)
 # The prereltag.txt is kept as a record in the distrib area
-# of the current pre-release version (currently not used explicitly 
+# of the current pre-release version (currently not used explicitly
 # anywhere for web pages/whatever).
-PRERELEASE_PREFIX=4\.4\.1~pre
-PRERELEASE_TAG=4.4.1~pre
+PRERELEASE_PREFIX=4\.4\.1-pre
```
zhenya1007: I certainly can: it's just text. :-)  However, isn't this project ultimately producing an Emacs package? If so, I submit that the relevant guide lines with respect to version numbers are  [the guide lines for Emacs package version numbers](https://www.gnu.org/software/emacs/manual/html_node/elisp/Packaging-Basics.html#Packaging-Basics).  My reading of those guide lines does not lead me to believe that "~" will "out of the box"  as part of an Emacs package version string ("out of the box" = "without having the user customize `version-separator` and/or `version-regexp-alist`").

zhenya1007:
Thank you for a thorough review.   I have pushed to my fork a change that (as far as i can tell) addresses all of your comments with the exception of the version string comment: we'll probably need one more "round trip" to resolve that.

Makefile.devel
erikmd: Hi @zhenya1007, thanks for your suggestion. Indeed, `4.4.1~pre` won't work out of the box with function `version-to-list`. But on the other hand, this issue with the tilde is not that critical currently, as the suffix `~pre` is not intended to be kept for PG releases: it was just added as a "SNAPSHOT" qualifier within the Git repo. However, according to the default value displayed in the documentation of `(describe-variable 'version-regexp-alist)`, a `-git` suffix would actually be preferable for that usage...
Thus, for the moment (to ensure the monotonicity of the version number) I propose you revert that change back to `4.4.1~pre` in the `Makefile.devel` file, and I shall apply this new suffix `-git` after the 4.5 version is released (namely, switching to something like `4.5.1-git`).

Makefile.devel
```diff
@@ -89,14 +89,14 @@ DEVELOPERS=\
 # PRERELEASE_PREFIX is used to match PRERELEASE_TAG in sed
 # line in tag target below, which edits $(DOWNLOADHTML)
 # The prereltag.txt is kept as a record in the distrib area
-# of the current pre-release version (currently not used explicitly 
+# of the current pre-release version (currently not used explicitly
 # anywhere for web pages/whatever).
-PRERELEASE_PREFIX=4\.4\.1~pre
-PRERELEASE_TAG=4.4.1~pre
+RERELEASE_PREFIX=4\.4\.1-pre
```
erikmd: Hi @zhenya1007,
Regarding these two lines, in addition to my suggestion in https://github.com/ProofGeneral/PG/pull/127#discussion_r94273448, there is a missing P in "RERELEASE_PREFIX" ;-)

Makefile.devel
```diff
@@ -309,51 +278,49 @@ distcompile: FORCE
...
-		rm $$f.old; \
-	   done) \
+#	if [ $(PRERELEASE_TAG) = $(VERSION) ]; then \
+#	  (cd $(HTMLDIR); \
+#	   for f in $(DOWNLOADHTMLS); do \
+#		mv $$f $$f.old; \
+#		sed -e 's|ProofGeneral\([emacselc-]*\)-$(PRERELEASE_PREFIX)......|ProofGeneral\1-$(PRERELEASE_TAG)|g' $$f.old > $$f; \
+#		rm $$f.old; \
+#	   done) \
 	fi
```
erikmd: This `fi` should be commented as well (the backslash in the previous line appears not to be sufficient).

Makefile.devel
zhenya1007: Thank you for considering the issue.  Your approach makes sense to me.

PR #128, opened 2016-12-07
A possible approach to generating autoloads.

zhenya1007:
This is essentially a variation on the approach suggested by @cpitclaudel.
The idea is that, since `package-install*' will call `update-directory-autoloads'
in the top-level directory of the package being installed exactly once
at installation time, code to "bootstrap" autoloads in sub-directories can be put into
a *.el file in the top directory of the package.


zhenya1007:
Is there something I can do to help with moving this along?  My motivation for suggesting this approach is that I think having something like my proposed `PG-autoloads-bootstrap.el` in the repository would make it fairly straightforward to write a MELPA recipe, and I would really like it if it was possible to install ProofGeneral "the same way you install every other MELPA package", rather than "by following some special installation instructions".  

One limitations of MELPA recipes (as far as I can tell) is that it isn't straightforward (and perhaps even not possible) to have a recipe that includes "here is a command to run" as part of the recipe.  (Naturally, nothing prevents me from opening a pull request against package-build to add such feature, but that's a different story. :-)).  Until MELPA recipes grow an ability to run commands, however, in order to write a MELPA recipe for ProofGeneral, II need either pre-generated autoloads or some code that will generate autoloads at the time the package is being installed.  This pull request takes the latter approach, while [realgud/realgud#84](https://github.com/realgud/realgud/issues/84) takes the former approach.  I don't have beliefs I hold dear either way: I just want ProofGeneral available on MELPA, hence this nag. ;-)

PG-autoloads-bootstrap.el
```diff
@@ -0,0 +1,18 @@
+;;;###autoload
```
cpitclaudel: Is ;;;###autoload documented to work before another set of comments? If not, I'd feel safer moving it to right before the function

PG-autoloads-bootstrap.el
```diff
@@ -0,0 +1,18 @@
...
+;;; Bootstrap ProofGeneral's autoloads.
+;; The code in package.el will not look in sub-directories of "the package directory"
+;; \(c.f. `package-generate-autoloads')
```
cpitclaudel: No need for the escaped parens here, nor the extra newline

PG-autoloads-bootstrap.el
```diff
@@ -0,0 +1,18 @@
...
+;; , so it needs a little help from the code below to find ProofGeneral's autoloads.
+(let* ((base-dir (file-name-directory
+                  (if load-in-progress load-file-name
+                    buffer-file-name)))
+       (autoloads-addendum (expand-file-name "PG-autoloads-addendum.el" base-dir)))
```
cpitclaudel: I'd prefer a name that clearly suggests that it's auto-generated :)

PG-autoloads-bootstrap.el
```diff
@@ -0,0 +1,18 @@
+;;;###autoload
+;;; Bootstrap ProofGeneral's autoloads.
+;; The code in package.el will not look in sub-directories of "the package directory"
+;; \(c.f. `package-generate-autoloads')
+;; , so it needs a little help from the code below to find ProofGeneral's autoloads.
+(let* ((base-dir (file-name-directory
```
cpitclaudel: Can we make this into a function, and just call the function from the autoload?

cpitclaudel: It looks like this code will run every time the file is loaded.  An `eval-when-compile` wrapper would probably help here :)

PG-autoloads-bootstrap.el
```diff
@@ -0,0 +1,18 @@
...
+                  (if load-in-progress load-file-name
+                    buffer-file-name)))
+       (autoloads-addendum (expand-file-name "PG-autoloads-addendum.el" base-dir)))
+  (unless (file-exists-p autoloads-addendum)
+    (let ((generated-autoload-file autoloads-addendum))
+      (dolist (name (directory-files-recursively base-dir "" t))
+        (when (file-directory-p name)
+          (update-directory-autoloads name)))))
+  ;; the load below is a wart, but it doesn't look I get full functionality if I just rely on autolads
```
cpitclaudel: I don't really understand what this part does.

cpitclaudel:
Thanks a bunch! I think the next step is to test this, to make sure that it works well for everyone.  Are you familiar with Cask? Can we come up with a Makefile recipe that builds a PG package?

cpitclaudel:
@zhenya1007 I'd love to merge this, but there are a few things that still need to be addressed.  Will you have time to have a look?

zhenya1007:
@cpitclaudel: I apologize for the delay.  I am going to work on it this weekend.

cpitclaudel:
No apologies needed at all! None of us has much time on their hands, so your help is very welcome, and greatly appreciated :) Thanks!

PG-autoloads-bootstrap.el
zhenya1007: It's a verbatim copy of the code you had [submitted to real-gud](https://github.com/realgud/realgud/issues/84).  It  walks the sub-directories of the directory in which this file is located, and adds the autoloads  found in each sub-directory to the `generated-autoload-file` (the call to `update-directory-autoloads` is in the dynamic scope of the let-binding, so the let-binding of `generated-autoload-file` is visible to the callee).

Or were you referring to something else?

zhenya1007:
@cpitclaudel: When I looked at this  again, I thought that generating autoloads at package installation time was a bit silly: the set of files from which the autoloads could possibly come from is already known at package build time, and won't change between package build time and package installation time (barring some unexpected intervention by the user).   The only thing generating autoloads at install time accomplishes is increase the amount of time needed to install the package.  For this reason, I adjusted things so that autoloads get generated at package build time, along with a small "shim" to pull the autoloads generated from sub-directories into the ProofGeneral-autoloads.el generated by Emacs's `package-generate-autoloads`. So far, I have not found any problems with using the  package built by the `pkg` target in Makefile.devel to work with Coq files.   I realize this update potentially amounts to more code for you to review, but I do think this is a cleaner approach.

To your question about:
> Are you familiar with Cask? Can we come up with a Makefile recipe that builds a PG package?

You already have a `pkg` target in Makefile.devel which produces a package.  For my ad-hoc testing I have been typing `make pkg GIT_RELEASENAME=HEAD EMACS=/path/to/emacs` to create a package, and then typing `M-x package-install-file` and pointing it at the tarball produced by the previous command.  Do you need a target like `pkg` but it a different makefile?  Or do you need a Cask file?  I am happy to work with you to include whatever you would like to see in the pull request, but I am not sure I understand what it is you would like to see. :-)

Finally, please let me know if you would like me to twiddle the commit history in some way (e.g, to start a new branch from the current state of `master`, and cherry-pick just  1ef6faf onto that branch, or somesuch).

Makefile.devel
```diff
@@ -511,23 +479,23 @@ releaseclean:
...
 ##     Do everything, but don't access CVS.  Just for
 ##     testing on non-live system, really.
-##	  
-## 
-fakereleaseall: 
-	$(MAKE) -f Makefile.devel release releaseclean NOCVS="no" 
+## FIXME: does this still make sense now that git is used version control?
+##
+fakereleaseall:
+	$(MAKE) -f Makefile.devel release releaseclean NOCVS="no"
```
cpitclaudel: To answer the fixme's question: no, I don't think so :)

PG-autoloads-bootstrap.el
cpitclaudel: Sorry, I meant the part below, not above :)

cpitclaudel:
Ok, thanks again for your work!  I agree with your approach :)  Can we go one step further, though, and just include the autoloads file in the repo?  IIUC we'll need this to get PG into MELPA, right?

That is, based on your suggestions, I think we should just make re-generating autoloads part of the default make target, but also make sure that they are in fact generated as a checked-in file, so which would live in the repo and thus be packagable by MELPA.

Can you rebase your code over the last master, so I can review the changes more easily? Right now github is showing me 41 commits, including background compilation changes :)

Re Cask files: parts of this makefile predate package.el; though it's convenient (and even more so with your changes and fixes, of course ^^), it won't work well with MELPA or ELPA, because both of these won't run our Makefile.  Instead, they'll just package together the files we tell them to package.  That's also why I think we need the autoloads file checked into the repository/

Thanks again for your help :) I think we're pretty close to getting all this working perfectly.

zhenya1007:
Hello again.

I took another look at this, and it looks like everything I need can actually be done by the rules in Makefile.devel.

The current patch I have is against current master, but I don't know how to update the branch on which this PR was open without force-pushing to it, so I will close this PR, and open a separate one (referencing this one).

PR #129, opened 2016-12-08
Keep going

hendriktews:
With option coq-compile-keep-going set, parallel compilation continues after the first error to compile as much as possible and to potentially report more than one error.

Need to

- ~~write CHANGES~~
- ~~write doc~~

before merging.

@cpitclaudel if you have time I would be happy to receive your feedback!


cpitclaudel:
Sorry for the delay; I've been busy with teaching.  This changeset looks great to me!

PR #132, opened 2016-12-12
Remove default absolute name from coq-prog-name (Fixes #76), but keep displaying…

### LINKED ISSUE #76: Why do we call proof-locate-executable on coq-prog-name (et al)?
This is the definition of `coq-prog-name`:

```
(defcustom coq-prog-name
  (proof-locate-executable "coqtop" t '("C:/Program Files/Coq/bin"))
```

Why do we need the proof-locate-executable part?

IIUC this computes an absolute path. The problem is that this prevents us from running PG in Tramp. 

Currently, running Coq remotely in Tramp only works when Coq isn't installed on the local machine, because coq-prog-name ends up being set to just `coqtop` in that case. But if Coq is installed, then we pass the full path of the local Coq binary to the remote server via tramp, and of course that doesn't work (the Coq binary isn't in the same location on the remote machine).

Just setting `coq-prog-name` to "coqtop" seems to work fine, though. Maybe we could do that?

Matafou:
… it when asking for it.

Fixing #76 trying to satisfy everyone: tramp users, windows users and users switching coq versions often.

Basically use "coqtop" everywhere except if coqtop is found outside windows path.
Show the absolute filename when asking the proof-prog-name interactivelt.

### DIFF
--- a/coq/coq-system.el
+++ b/coq/coq-system.el
@@ -30,7 +30,8 @@ On Windows you might need something like:
   :group 'coq)
 
 (defcustom coq-prog-name
-  (proof-locate-executable "coqtop" t '("C:/Program Files/Coq/bin"))
+  (if (executable-find "coqtop") "coqtop"
+    (proof-locate-executable "coqtop" t '("C:/Program Files/Coq/bin")))
   "*Name of program to run as Coq. See `proof-prog-name', set from this.
 On Windows with latest Coq package you might need something like:
    C:/Program Files/Coq/bin/coqtop.opt.exe
--- a/coq/coq-system.el
+++ b/coq/coq-system.el
@@ -43,13 +43,15 @@ See also `coq-prog-env' to adjust the environment."
   :group 'coq)
 
 (defcustom coq-dependency-analyzer
-  (proof-locate-executable "coqdep" t '("C:/Program Files/Coq/bin"))
+  (if (executable-find "coqdep") "coqdep"
+    (proof-locate-executable "coqdep" t '("C:/Program Files/Coq/bin")))
   "Command to invoke coqdep."
   :type 'string
   :group 'coq)
 
 (defcustom coq-compiler
-  (proof-locate-executable "coqc" t '("C:/Program Files/Coq/bin"))
+  (if (executable-find "coqc") "coqc"
+    (proof-locate-executable "coqc" t '("C:/Program Files/Coq/bin")))
   "Command to invoke the coq compiler."
   :type 'string
   :group 'coq)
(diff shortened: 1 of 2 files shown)


hendriktews:
I would suggest to change the calls for coq-dependency-analyzer and coq-compiler in the same way as for coq-prog-name.

hendriktews:
... and I would suggest to include ``(fixes #76)'' in the commit message to save one click when this PR gets merged

Matafou:
Done.

Matafou:
I will merge this if no more comment.

patcaslm:
This commit seems to be causing problems on MacOs. Since I did a git pull a few days ago, Proof General cannot find the coq binaries anymore.

cpitclaudel:
@patcaslm Can you give more details about your config? Where are your coq binaries, which errors do you get, etc?

patcaslm:
I did more digging and now I am not so sure the problem is because of the changes to coq-system.el. I evaluated the buffer with coq-system.el and then looked at the variables coq-prog-name, coc-dependency-analyzer, and coq-compiler and their values were "coqtop", "coqdep", and, respectively, "coqc", which seems just right. These binaries are in the PATH variable and I can invoke them from the terminal. The problem is that Proof General, for some reason, cannot find them. I could see errors like the following in the logs:

/bin/bash: coqtop: command not found
proof-activate-scripting: Searching for program: No such file or directory, coqtop

This doesn't make sense to me. I can execute coqtop with bash, in the terminal. Any ideas?

cpitclaudel:
How are you starting emacs? It could be that Emacs doesn't inherit your path properly.

Matafou:
Hi,
It looks like you don"t have the right PATH in your X session. 

Try M-! echo $PATH

to see what is the path of the shell you call from emacs (path inherited from the X session).

Fixing this needs to understand how your X session manager (gdm, xdm etc) reads you shell config files.

Matafou:
mmh I read too fast. If the variable a set correctly then it is not X session problem.

cpitclaudel:
My personal bet would be that Emacs is being started from the dock, not from the command line

Matafou:
Probably yes. But if the coq-prog-name is set correctly it means that the path is visible somewhere but not everywhere.

@patcaslm Can you please compare the result of

M-! echo $PATH
and the value of variable exec-path?

Are you using aquamacs?

patcaslm:
Hi everyone,

I rebooted and everything works well now.

The problem was indeed with the PATH that emacs was seeing (both Aquamacs and Emacs, and it didn't matter if they were started from the Dock or from a terminal -- weird). A few days ago I installed Coq 8.6 from Opam (until then I used to use Macports). At the same time I also updated Proof General. But the problem seems to have been caused by the new installation of Coq.

Thanks for your help. You put me on the right track 👍

PR #137, opened 2016-12-19
Use new PG image for Help toolbar 

psteckler:


psteckler:
The 6 commits shown here are a red herring. The only difference between this branch and the PG master branch should be the 2 help icon files.

images/epg-help.xpm
```diff
@@ -1,248 +1,257 @@
 /* XPM */
-static char * epg_help_xpm[] = {
-"24 24 221 2",
+static char * help_xpm[] = {
```
cpitclaudel: Should this be called epg_help_xmp intead of help_xmp? (Does it not make a difference?)

cpitclaudel:
Thanks! I can just squash and merge to remove the spurious commits.

images/epg-help.xpm
psteckler: Probably it should be changed (not sure if it makes a difference). 

The name is an artifact of how I created the file.

psteckler: I'll make the change in my branch and push.

psteckler: Done.

cpitclaudel:
Thanks! One more thing: the original was a square. Did that matter?

psteckler:
Hmmm, I thought I'd asked gimp to make it 24x24. Probably when I set one dimension, it chose the other one. Let's see if I can fix that.

psteckler:
Because the image was taller than wide, I cropped off the bottom to get a square image, then shrunk it down to 24x24, the old size.

Issue #154, opened 2017-02-20
Feature request: report missing Qed. somehow


Blaisorblade:
With Coq 8.5, ProofGeneral appears to accept the following script (minimized from a 1000-line proof script).

```
Lemma a1 : 1 + 1 = 2.
auto.

Lemma a2 : 1 + 2 = 3.

Lemma a3 : 1 + 3 = 4.
  auto.
Qed.
```

However, `coqc` rejects it:
```
$ coqc Foo.v
Error: There are pending proofs
```

@JasonGross mentioned that the pending proofs can be discovered with `End` (and `End a3` indeed shows the missing proofs), or `Show Proof.` (but that only shows `?Goal` with Coq 8.5).

But at least I (a Coq newbie) can't honestly tell from the ProofGeneral interface — when I step to the end of the script, it shows an empty `*goals*` window, at least sometimes.
So much that the 1000-line script got committed by our resident Coq expert (not me).

OTOH, if I resume the outer proofs with `auto. Qed. Qed.` I get a warning:

> Warning: Nested proofs are deprecated and will stop working in the next Coq version

suggesting the problem might fix itself soon (though this is debated https://sympa.inria.fr/sympa/arc/coq-club/2015-08/msg00031.html).

Quoting @JasonGross wrote that https://github.com/ProofGeneral/PG/issues/37#issuecomment-281146280:
> Report a separate issue for the Qed; the comment issue described here is a bug in PG; PG parses the document differently from coqtop and coqc, and this causes unexpected behavior. Your issue is that PG doesn't tell you if your file is unfinished (has unterminated proofs or open sections), but coqtop acts the same way as PG here; it's not clear when or how PG should give you this information (and you can discover it at any time by seeing what happens if you do Show Proof. or End foo.; no need to bisect), so it's more properly a feature request. I'd actually suggest making a feature request on the Coq bug tracker, asking for the error message to tell you what line the proof started on, and possibly the theorem name.

Blaisorblade:
>  I'd actually suggest making a feature request on the Coq bug tracker, asking for the error message to tell you what line the proof started on, and possibly the theorem name.

I found this reported as https://coq.inria.fr/bugs/show_bug.cgi?id=5275.

Matafou:
Hi,
To be honest I don't see how and why Proofgeneral should complain about an
unterminated proof. After all most of the time I do have an unterminated
proof at the end of my buffer, that is the whole point of proof interfaces:
[...]

PR #156, opened 2017-02-23
Fixing #154.

Matafou:
This an attempt at fixing #154.
It seems to work, I would like to discuss the interaction with @hendriktews 's proof-tree, of which I mimick the way of inserting invisible commands in the action-list (nice by the way).

Hendrik, can you have a look please? I wonder if I should insert my commands before or after yours in proof-shell-exec-loop. Both are "Show." but mine is not no-response.

We could also merge both insertion in a single generic (proof-shell-invisible-commands-after-advancing-command). Or we may not bother since xml-pg has a different approach anyway.

### DIFF
--- a/coq/coq.el
+++ b/coq/coq.el
@@ -1218,6 +1218,15 @@ flag Printing All set."
     (remove-hook 'proof-assert-command-hook 'coq-adapt-printing-width)
     (remove-hook 'proof-retract-command-hook 'coq-reset-printing-width)))
 
+;; In case of nested proofs (which are announced as obsolete in future versions
+;; of coq) Coq does not show the goals of enclosing proof when closing a nested
+;; proof. This is coq's proof-shell-empty-action-list-command function which
+;; inserts a "Show" if the last command of an action list is a save command and
+;; there is more than one open proof before that save.
+(defun coq-empty-action-list-command (cmd)
+  (when (and (string-match coq-save-command-regexp-strict cmd)
+             (> (length coq-last-but-one-proofstack) 1))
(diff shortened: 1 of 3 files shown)


Matafou:
Well. I just checked How prooftree behaves, but prooftree does not support nested proofs. So there is no interaction between this patch and prooftree.
If someone wants to have a look anyway, I will merge in 1 or 2 days.

Blaisorblade:
Thanks for the patch! Makes lots of sense to me (haven't tested yet).
Hopefully stupid question: of course calling `Show` when the proof state is already shown is redundant. Is `Show.` ever slow enough to matter in realistic uses? I'm guessing not, but what do I know.

Matafou:
Show is sent only when closing a nested proof. It happens rarely.

hendriktews:
> Hendrik, can you have a look please? I wonder if I should insert my
> commands before or after yours in proof-shell-exec-loop. Both are
> "Show." but mine is not no-response.

The prooftree show commands at this point are used to retrieve
the full text of new proof goals as well as the updated text of
old goals after existentials have been instantiated (where the
latter does not work with coq >= 8.5). Therefore the order does
not matter, as long as the commands coming from
proof-shell-empty-action-list-command do not change the proof
state.

> We could also merge both insertion in a single generic
> (proof-shell-invisible-commands-after-advancing-command). Or we may
> not bother since xml-pg has a different approach anyway.

I would also not bother... But I see more and more hacks to
modify proof-action-list. It would be nice to have a clean
interface for all this...

Hendrik

PR #157, opened 2017-02-26
[WIP] ELPA/MELPA support

cpitclaudel:
I finally found the time to look into ELPA/MELPA support.  Here's a brief summary of the issues that popped up:

* The current Makefile uses a fresh Emacs to compile each file.  On the other hand, package.el compiles all files in the same Emacs instance (the current one).  This means that the common pattern of running `(proof-ready-for-assistant …)`  in an `(eval-when-compile …)` wrapper doesn't work (it used to work because there was only one such for per file and each file was compiled in a fresh Emacs).

* There were uses of defpacustom and defpgdefault in files of the coq/ hierarchy that were `require`d by other files.  This only works *after* running `(proof-ready-for-assistant)` — without that, running `(defpacustom x …)` at load time just creates a variable called `nil-x`, because the proof assistant name isn't set.  But we can't run proof-ready-for-assistant, so I replaced these by `defcustom`s and `setq-default`.

* Some forms were `eval-and-compile`-d, without justification (and some had `;; FIXME Why?` near them; this includes `coq-mode`.  One comment mentioned ensuring that `…-mode-map` would be defined at compile time, but why would we want that?

* Autoloads are handled by `proof-site` for now (it loads `generic/pg-autoloads`).  How do we refresh that file? This connects to #128.

* PG had a pervasive assumption that noninteractive is set iif we're currently byte-compiling.  That assumption is wrong.

* There were a bunch of `(defvar … nil)` to silence the byte-compiler's warnings.  Isn't `(defvar …)` the proper form (no initializer)?

* I selected a subset of all directories for packaging.  I excluded the following: `bin etc obsolete previous-art  "acl2" "ccc" "hol-light" "hol98" "isar" "lego" "phox" "twelf"` (all the provers listed in there are marked as deprecated, except for isar).

* These changes add some compilation warnings, due to the byte-compiler not finding some symbols.  The normal way to do fix those be to `(require)` the appropriate files.  We can't do that, unfortunately, so instead we need to `(defvar …)` the relevant variables.  Help with this would be welcome.

Hopefull @Matafou and @erikmd can have a look (@hendriktews, if you have time to look at the compilation parts, that would be great too) :) To test, please clone https://github.com/cpitclaudel/melpa/, and run `make clean-sandbox recipes/proof-general sandbox`.  This opens a fresh Emacs in which you can run M-x package-install proof-general, and then open Coq files and see that everything works.  @strub, if you have time to test easycrypt support, that would be great too.

A careful review of the diffs would also be very welcome :) Thanks!

### DIFF
--- a/proof-general-pkg.el
+++ b/proof-general-pkg.el
@@ -1,0 +1,2 @@
+(define-package "proof-general" "4.4.1~pre" "A generic front-end for proof assistants (interactive theorem provers)"
+  '((emacs "24.4")))
--- a/coq/coq-abbrev.el
+++ b/coq/coq-abbrev.el
@@ -71,7 +71,7 @@
     ;; DA: how about above, just temporarily disable saving?
     (message "Coq default abbrevs loaded")))
 
-(unless noninteractive
+(unless (or noninteractive (bound-and-true-p byte-compile-current-file))
   (coq-install-abbrevs))
 ;;;;;
 
--- a/coq/coq-abbrev.el
+++ b/coq/coq-abbrev.el
@@ -98,7 +98,7 @@ It was constructed with `proof-defstringset-fn'.")
 
 ;; The coq menu partly built from tables
 
-;; Common part (scrit, response and goals buffers)
+;; Common part (script, response and goals buffers)
 (defconst coq-menu-common-entries
   `(
     ["Toggle 3 Windows Mode" proof-three-window-toggle
@@ -300,7 +300,7 @@ It was constructed with `proof-defstringset-fn'.")
     ["ML4PG" (coq-activate-ml4pg) :help "Activates ML4PG: machine-learning methods for Proof General"]
     ))
 
-(defpgdefault menu-entries
+(setq-default coq-menu-entries
   (append coq-menu-common-entries
   `(
     ""
@@ -338,12 +338,10 @@ It was constructed with `proof-defstringset-fn'.")
      ["help" coq-local-vars-list-show-doc t]
      ["Compile" coq-Compile t]))))
 
-(defpgdefault help-menu-entries
+(setq-default coq-help-menu-entries
   '(["help on setting prog name persistently for a file" 
      coq-local-vars-list-show-doc t]))
 
-(defpgdefault other-buffers-menu-entries coq-menu-common-entries)
-
-
+(setq-default coq-other-buffers-menu-entries coq-menu-common-entries)
 
 (provide 'coq-abbrev)
--- a/coq/coq-autotest.el
+++ b/coq/coq-autotest.el
@@ -16,7 +16,7 @@
   (defvar coq-compile-before-require nil))
 
 
-(unless noninteractive
+(unless (bound-and-true-p byte-compile-current-file)
 
   (pg-autotest start 'debug)
 
--- a/coq/coq-autotest.el
+++ b/coq/coq-autotest.el
@@ -10,11 +10,8 @@
 
 (require 'pg-autotest)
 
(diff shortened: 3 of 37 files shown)


psteckler:
One of the things I did for the XML code was to change uses of the 'cl library to 'cl-lib, per the guidelines at: https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html.

I wonder if that should be done in the master branch.

One thing I haven't done that's mentioned in those guidelines is to use 1 to 3 prefixes for naming variables and functions. The PG source code uses many prefixes.

psteckler:
Many of the changes in this pull request apply to the XML code as well. If the PR is merged, I can merge it back, although I think some of the changes won't apply.

cpitclaudel:
> One of the things I did for the XML code was to change uses of the 'cl library to 'cl-lib, per the guidelines at: https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html.

It should be done, but probably not as part of this PR, I think

psteckler:
I can verify that the Melpa package install works (on Emacs 24.5), and I was able to process a couple of Coq files after that.

cpitclaudel:
Cool, thanks for testing.  @tchajed, do you think you could give it a go, too?

proof-general-pkg.el
```diff
@@ -0,0 +1,2 @@
+(define-package "proof-general" "4.4.1~pre" "A generic front-end for proof assistants (interactive theorem provers)"
+  '((emacs "24.4")))
```
marsam: I think this can be 24.3

tchajed:
It seems to work for me as well, on OS X 10.11 and Emacs 25.1.1.

proof-general-pkg.el
cpitclaudel: Thanks.  I think that's probably true, but I haven't tested it there, and I don't think we're regularly doing that, so I'd rather be conservative.  Do you know many users of 24.3?

cpitclaudel:
@tchajed Thanks!

marsam:
Awesome work. Tested on Emacs git and everything seems working fine.

+ About mmm.  Checking the [contrib/mmm](https://github.com/ProofGeneral/PG/commits/master/contrib/mmm) logs most of the commits are imports from the mmm repository, only   https://github.com/ProofGeneral/PG/commit/9429d7e54d4674a43975577927b0e3af0d7f88b5   tries to fix Emacs 23.3 support. The official mmm es well maintained, so I   think this is already fixed. If we plan to make it available in melpa, we   should depend on the official mmm.
  
+ (minor) I think melpa expects a package to use a prefix as [elisp convention](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html) so the packages under `/lib` should be added a prefix `proof-`?

cpitclaudel:
Thanks a lot for testing.  I agree about mmm, though that sounds like one more large piece of work, and I don't know which parts of PG depend on it.  The proof prefix is a good point too, though maybe we can get a waiver :)

proof-general-pkg.el
marsam: > Do you know many users of 24.3?

I don't have any stats, but is the default version in Ubuntu 14.04 LTS which is supported until 2019-04.

cpitclaudel: Good point, thanks :) I'll test there then.

rgrinberg:
@cpitclaudel awesome work. Just curious, how close are you to getting PG into melpa? The make step has to be removed entirely right?

PR #162, opened 2017-03-03
Refreshing goal when Set Printing xxx.

Matafou:


cpitclaudel:
LGTM.  Thanks!

erikmd:
OK so I'm going to squash it.

psteckler:
I'm not certain whether/how to apply this change to PG/xml, which isn't maintaining the last-but-one-proofstack variable, and the role of the proof-action-list is different there: items on that list are handed to the transaction queue, rather than being sent directly to Coq. Also, selecting an option like "Set Printing All" from the menus puts an item directly in the transaction queue, bypassing the action list (because it was easier for me to write that, which may not be a good reason).

Matafou:
It is an interesting question. It seems unharmful for printing options. But what about options that really change the behaviour of Coq?

We are already supposed to forbid dangerous options to be send to coq directly, but this way of bypassing the action-list makes it even more important.

PR #166, opened 2017-03-05
Generate autoloads for sub-directories.

zhenya1007:
Create an autoloads file that collects autoloads from Elisp files in
sub-directories of ProofGeneral's  top-level directory.  The reason
for generating this file is that the standard Emacs package
installation procedure will only inspect the Elisp files in
ProofGeneral's top-level directory, but will not inspect the
sub-directories (c.f `package-generate-autoloads' in the standard
Emacs distribution).

The `pkg' target of the Makefile now creates a small "shim" to cause
the autoloads collected from sub-directories (described in the
preceeding paragraph) to be loaded whenever ProofGeneral-autoloads.el
file (which is generated by `package-generate-autoloads') is loaded.

Finally, this change adds a couple of autoload "cookies" which were
needed (at least in my testing) so that Coq mode can be used
immediately upon package installation (i.e., in the same Emacs session
in which `package-install <a_location_of_ProofGeneral_package>' was
run.  Without those cookies, I was getting an error complaining that
the function `coq-mode-config' was undefined.  Similar "cookies" may
be needed for other proof assistants (I only use Coq, so I am choosing
to stay conservative in my changes).

Note: the discussion took place under PR 128, but I couldn't figure out how to get Git to bring the branch associated with PR 128 into a reasonable state without force-pushing.


zhenya1007:
Any chance of merging this?

zhenya1007:
Is there something I can do to help move this along?

erikmd:
@zhenya1007 Thanks for proposing this PR! and sorry for the delay of reply. 

I've just reviewed your patch and re-read #128 where the discussion had start, and it appears that this PR #166 is unneeded to get proper `package.el` and MELPA support w.r.t. autoloads. To be more precise:

- As @cpitclaudel had mentioned in https://github.com/ProofGeneral/PG/issues/222#issuecomment-412596600, *the second part of your PR* to ensure `generic/proof-site.el` is loaded is redundant with the `pg-init.el` file introduced in #157.
- *The first part of your PR* that collects autoloads at build time in a new file `PG-generated-autoloads-from-subdirs.el` doesn't address Clément's remark https://github.com/ProofGeneral/PG/pull/128#issuecomment-277601314 as one such file (if it were necessary) should be already added to the repo because MELPA's machinery doesn't rely on `make`.
- This file `PG-generated-autoloads-from-subdirs.el` is not useful because all autoloads are in `generic` and `lib` subdirs IINM, which are already considered when generating `generic/proof-autoloads.el`, which is itself required by `generic/proof-site.el`.
- Finally the autoload cookie before `coq-mode-config` doesn't seem necessary as it is not an interactive function and I didn't experience any error like "coq-mode-config was undefined" with the tests I performed with a local MELPA installation with my `prepare-melpa` branch.

So I'm going to close this PR. Anyway thanks @zhenya1007 for contributing to these discussions!

Issue #158, opened 2017-02-27
Can't show comment after hiding it


psteckler:
After running Coq on this script, if I use the context menu to hide the second comment here:

`(* a not-so-interesting theorem *)`
`Goal True.`
  `(* this one should be easy to prove, eh? *)`
  `auto.`
`Qed.`  

then I can again choose that menu to show it again. But if I hide the first comment, I don't seem to be able to get the context menu again in order to show it.

psteckler:
The problem is that the span for the script item and the help span are co-extensive, so that when the script item span is made invisible, the help span and its context menu are inaccessible.

A solution would be to make the help span, constructed in `pg-set-span-helphighlights`, to extend one character beyond the script item span on both ends (with exceptions for the start and end of the buffer).

Shall I create a PR for that?

erikmd:
Hi @cpitclaudel @Matafou,
I've just tried reproducing this issue #158, but was blocked with this question:

What is the key-binding/mouse-binding necessary to make the context menu appear?
is it disabled by default?

Indeed this could be useful to know before merging #169 (but it is not urgent anyway!)

erikmd:
Actually it seems that I currently have a local issue with my main config, since when I run a fresh Emacs from the melpa repo doing `make clean-packages clean-sandbox recipes/proof-general sandbox`, that allows me to reproduce this issue with right-click, i.e., `<mouse-3>` :)

erikmd:
> What is the key-binding/mouse-binding necessary to make the context menu appear?
> is it disabled by default?

OK so I figured out the source of that small issue: when using both PG master and company-coq, the context menu doesn't show up when one right-click on a Coq phrase.
@cpitclaudel I'll open a small issue on your repo. 

Anyway when company-coq is not loaded, I can reproduce #158 and PR #169 indeed fixes this so I can merge it.

cpitclaudel:
Oh I remember now. I don't think it's fixable :'(

PR #169, opened 2017-03-09
extend helpspan, issue #158

psteckler:
With this change, the help context menu is accessible on either side of the hidden comment.


erikmd:
Merging now (using PG master without company-coq, one can indeed reproduce #158 and fix it using this).

PR #170, opened 2017-03-14
[doc]: add documentation for the EasyCrypt mode

strub:


cpitclaudel:
Thanks. Please ping us if something easy like this stays open for so long :/

PR #178, opened 2017-04-18
Add file contrib/mmm/.nosearch to help feature extraction tools

tarsius:
This change is intended as an intermediate step toward the removal
of the bundled copy of `mmm` package.  Even with this change, PG
continues to use the bundled version, unless `mmm-auto` is already
loaded at the time PG first loads `proof-auxmodes`.

The benefit of this change is that tools that extract the features
provided and required by a package, such as those used to maintain
the Emacsmirror, will no longer be tricked into believing that
`mmm` is part of PG.

Eventually the bundled `mmm` copy should be removed completely, as I
suggested in #171, and as was already on the roadmap before I did so.


cpitclaudel:
Thanks! Sorry for the delays.

tarsius:
Thanks for the merge! ;-)

Issue #177, opened 2017-04-15
/usr/bin/proofgeneral is not executable by default?


wilbowma:
When I build using `make PREFIX=/usr install`, the executable for `proofgeneral` is not make executable. Of course, I could change this myself, but it seems like something `make install` should do. Is this by design or am I missing something?

cpitclaudel:
I don't think many people use this installation medium any more; instead, we'd like to move to making Proof General a proper Emacs package (you can help! https://github.com/ProofGeneral/PG/pull/157).  That could be why you're running into this issue. (I'd be happy to merge a fix for the problem you describe, though, of course :)

wilbowma:
Aha! I'm maintaining the ArchLinux package for proofgeneral and a user reported this as a bug: https://aur.archlinux.org/packages/proofgeneral. I'll try to send a PR

hendriktews:
The proofgeneral shell script in the Proof General repo is
leftover from ancient days. I don't think you want to install
this for your users. For Debian I substituted this with a simple
script that only starts emacs:

    #!/bin/sh

    if emacs --version | grep -q 'Emacs 2[45]' ; then 
        emacs -f proof-splash-display-screen $*
    else 
        if emacs25 --version | grep -q 'Emacs 2[45]' ; then 
            emacs25 -f proof-splash-display-screen $*
        else 
            emacs24 -f proof-splash-display-screen $*
        fi
    fi

Hendrik

wilbowma:
What's the downside of installing it? It seems to work just fine.

wilbowma:
I'm pretty confused; it looks like `bin/proofgeneral` is executable in the repo and the `Makefile` uses `cp -p`. It seems like it should be installed executable by `make install`.

hendriktews:
> What's the downside of installing it? It seems to work just fine.

I don't know any real reason why you should not install it. It's
just that it contains quite a bit of unnecessary code that is not
maintained anymore ...

hendriktews:
> It seems like it should be installed executable by make
> install.

IMHO, instead of fixing make install, we should remove
bin/proofgeneral from the repository.

Hendrik

Matafou:
I agree with @hendriktews we should remove it.

PR #180, opened 2017-04-24
proofgeneral script -- OS X fixes

gregory-nisbet:
So, I made a few changes to the script. Here's what they are and the motivation.

- readlink_canonicalize function

OS X's readlink does not have a -f option. It is possible to hack together a pure POSIX solution using `pwd -P`, but it's not really intuitive. So we call perl if readlink doesn't support the `-f` option.

- quoting and `$(..)`

There are some cosmetic changes related to placating shellcheck ... mostly quoting variables and replacing `` `..` `` with `$(..)`.

- Using `/Applications/.../Emacs` or `/Applications/.../Aquamacs` on OS X

I think a significant number of people on OS X use the Emacs for OS X cask available from Homebrew. Or Aquamacs. I added some logic to default to pick Emacs for OS X if it's available, then try Aquamacs, then fall back to `which emacs` as before.

Stuff that I'm considering doing but haven't done yet.

- lowercase all variables that aren't intended to be inherited from the environment.
- rewrite this thing in Perl. There are already some Perl scripts in the repo, and Perl makes certain things like getting the directory of the file that's currently executing reliable ( `"$(basename "$0")"` is not guaranteed to work.).

### DIFF
--- a/bin/proofgeneral
+++ b/bin/proofgeneral
@@ -20,16 +20,31 @@
 # NB: no trailing backslash here!
 # On Mac, maybe:
 # /Applications/Emacs.app/Contents/MacOS/Emacs/site-lisp/ProofGeneral
-PGHOMEDEFAULT=$HOME/ProofGeneral
 
-NAME=`basename $0`
+
+EMACS_FOR_OSX_EMACS="/Applications/Emacs.app/Contents/MacOS/Emacs"
+AQUAMACS_EMACS="/Applications/Aquamacs.app/Contents/MacOS/Aquamacs"
+
+readlink_canonicalize() {
+    # detect whether readlink understands the -f option (e.g. GNU readlink)
+    if readlink -f . 1>/dev/null 2>&1; then
+        readlink -n -f "$@"
+    # otherwise fall back to perl
+    else
+        perl -MCwd -e 'print Cwd::abs_path($ARGV[0])' "$@"
+    fi
+}
+
+
+PGHOMEDEFAULT="${HOME}/ProofGeneral"
+NAME="$(basename "$0")"
 
 HELP="Usage: proofgeneral [OPTION] [FILE]...
 Launches Emacs Proof General, editing the proof script FILE.
 
 Options:
   --emacs <EMACS>     startup Proof General with emacs binary <EMACS>
-  --pghome <PGDIR>    startup Proof General from directory PGDIR
+  --pghome <PGHOME>    startup Proof General from directory PGHOME
   -h, --help          show this help and exit
   -v, --version       output version information and exit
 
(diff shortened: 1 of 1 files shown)


Matafou:
As discussed elsewhere some people think we should just remove this script completely. Why is it needed?

Concerning defaulting to homebrew I hear a lot of people prefer aquamacs over homebrew emacs, could it be possible to have both working out of the box?

gregory-nisbet:
Done, the script should detect Emacs for OS X and Aquamacs. It does this by checking for `Emacs` and `Aquamacs` executables in `/Applications/...`. 

I like the script. It makes it easier to mess around with my emacs configuration while having a "clean" configuration just for Proof General that's less likely to get broken.

erikmd:
Thanks @gregory-nisbet for your proposal!, but as mentioned @Matafou there is ongoing discussion to remove this shell script that was (intended to be) installed by "make install". IMHO it is unnecessary to maintain a 100-line script to merely do `emacs -f proof-splash-display-screen "$@"`.
I would prefer to keep it simple, and decide between:
Option 1: compress that script to the bare minimum (e.g. like @hendriktews's suggestion https://github.com/ProofGeneral/PG/issues/177#issuecomment-294338161)
Option 2: remove it from the PG repo, leaving the choice to include such a script at the discretion of the maintainers of distributions.
Also, note that there is a pending PR #157 that aims at making PG a proper Emacs package for next release:  after this, the bin/proofgeneral script will be really unnecessary.

Personally I would rather vote for Option 2 and close #177 and #180 accordingly.
@cpitclaudel, your opinion?

cpitclaudel:
+1 on @erikmd's answer.  I really want to move towards #157

erikmd:
Done in master with 1a18e33658645a81225c56b5d4f4a4b89434d301, closing this PR as no more relevant.

hendriktews:
> Done in master with 1a18e33, closing this PR as no more relevant.

Great, thanks!

H.

Matafou:
Sorry but I think you are a bit too fast here. Gregory's answer makes me think that we should maybe poll pg users about the need for this script.

gregory-nisbet:
I'm mostly a coqide user right now, but I'm trying to switch to proof general. (I'm working my way through Software Foundations on my own at the moment.) My motivation for proposing these changes to the script was thinking back to an SF Types and Theorems meetup. Some people, including me, struggled to get proof general to work because they weren't elisp hackers in the past. I'd love to be able to point people at a repo and say clone this and run this script and you will get a graphical proof-general that's independent of your emacs setup.

cpitclaudel:
I'm happy to re-instate a simpler, smaller script that just runs `emacs -Q` with a temporary home, if that's useful.  In fact, MELPA has this (with `make sandbox`, you get a clean Emacs in which you can install anything).

But, I'd much rather fix the issues that cause actual problems.  In the idea world, here's what should happen: `M-x list-packages`, `M-x package-install RET proof-general RET`, and then just open a `.v` file.  If you're not an experienced ELisp hacker, nothing in your config should trip up PG.

cpitclaudel:
> Some people, including me, struggled to get proof general to work because they weren't elisp hackers in the past.

Do you have a concrete example of an issue that people ran into ? We'd love to fix those :)

PR #189, opened 2017-06-20
Fix easycrypt automode regexp

marsam:
Without the string-end (`\\'`) it will match all file names which contain `.eca?`.


cpitclaudel:
Thanks a lot

PR #191, opened 2017-06-30
Formatting fix for proof-layout-windows documentation

AndreasLoow:
`make html` didn't work on my machine so I'm not sure how the html looks like, but the pdf generated by `make pdf` seems ok.


Matafou:
Thank!

PR #196, opened 2017-08-12
Cleanup PR to prepare the integration of PG/xml

erikmd:
This PR is a proposal to remove legacy files from the main PG repo (that won't be part of PG/xml) to facilitate the integration of PG/xml (the diff will be more concise). A commit also updates the AUTHORS file.

Cc @cpitclaudel @Matafou @psteckler


hendriktews:
Erik Martin-Dorel <notifications@github.com> writes:

> This PR is a proposal to remove legacy files from the main PG repo

Very good! I am happy to see that somebody is brave enough to
clean up all this old stuff.

Because github is not so clear about it, let me summarize: You
completely deleted the following subdirectories
- acl2
- ccc
- easycrypt
- hol
- light
- hol98
- isar
- lego
- obsolete
- pghaskell
- pgocaml
- pgshell
- phox
- twelf
- etc/hol-light
- etc/isar
- etc/lego
- etc/trac

Are you sure about easycrypt? ASFAIK, Clement merged some PR for
easycrypt not so long ago. My impression was that Proof General
is currently used for Coq and easycrypt.

Have you thought about the documentation? When all the code is
gone, it would be good to delete the corresponding chapters in
the user manual.

Hendrik

erikmd:
Hi Hendrik! thanks for your comment.

I have to precise that all the subfolders removed in this PR (including easycrypt) have already been removed in [Paul's fork](https://github.com/psteckler/ProofGeneral) because PG/xml does not support anymore the REPL architecture of legacy PG
(e.g., there is no more `generic/proof-shell.el` in PG/xml, while easycrypt requires it) 

So the usefulness of this PR is especially to remove the legacy code from the master branch before integrating PG/xml so the PG/xml patch won't be cluttered with all these removals.

Of course, we'll keep all the legacy codebase in a separate "legacy" branch so all users can continue using easycrypt or PG's support for other provers from that legacy branch.

> Have you thought about the documentation? When all the code is gone, it would be good to delete the corresponding chapters in the user manual.

Absolutely, we'll need to update the documentation accordingly. Though I believe it is unnecessary to update it immediately in this PR, as we'll be able to do it once PG/xml is merged.

craff:
I just updated phox ... and it now works and have a pull request

erikmd:
Rebased branch `rm-legacy` (#196) on `async`.

erikmd:
New rebase of `rm-legacy` on `async`. Will merge in `async` as soon as Travis is OK.

PR #200, opened 2017-09-21
phox is back

craff:
was in fact quite easy ... thanks to easy config ...


phox/phox.el
```diff
@@ -1,231 +1,141 @@
...
-; ne sont pas pris en compte.  Les prefixes de moins de 3 caract�res ne
-; sont pas non plus pris en compte.
-
-; (set-variable 'phox-completion-table
-(defpgdefault completion-table
-  (append phox-top-keywords phox-proof-keywords)
-)
+
+
+;; code for sisplaying unicode borrowed from
```
siddharthist: Typo here

erikmd:
Hi @craff! Thanks a lot for your PR! and apologies for the delay of reply :/

As we're going to submit PG `master` (based on proof-shell = REPL-based) on MELPA (see #222), it was important for us to integrate your PR beforehand, so that the `proof-general` package will essentially contain PG support for `coq` REPL, `easycrypt`, and `phox`.
So I'm going to merge #200 right now in `master` (then fix the typo mentioned by @siddharthist later on)

Thanks again!

PR #207, opened 2017-10-24
Make coq-prog-args safe when list of strings.

SkySkimmer:
This is similar to the builtins `compile-command`/`compilation-read-command`.

### DIFF
--- a/coq/coq-system.el
+++ b/coq/coq-system.el
@@ -22,6 +22,15 @@
   (defvar coq-prog-args)
   (defvar coq-debug))
 
+;; Arbitrary arguments can already be passed through _CoqProject.
+;; However this is not true for all assistants, so we don't modify the
+;; (defpgcustom prog-args) declaration.
+(defun coq--string-list-p (obj)
+  "Determine if OBJ is a list of strings."
+  (or (null obj) (and (consp obj) (stringp (car obj)) (coq--string-list-p (cdr obj)))))
+
+(put 'coq-prog-args 'safe-local-variable #'coq--string-list-p)
+
 (defcustom coq-prog-env nil
   "List of environment settings d to pass to Coq process.
 On Windows you might need something like:


SkySkimmer:
I looked for `defcustom.*-prog-name` for things to change. Tell me if any defvars should be added.

psteckler:
The change to `coq-system.el` should also find its way to PG/xml, if this is merged.

generic/proof-useropts.el
```diff
@@ -255,7 +255,12 @@ files which are out of date with respect to the loaded buffers!"
 
 (defcustom proof-prog-name-ask
   nil
```
Matafou: Should we set it to t by default? If you think yes please change the PR.

psteckler: That would be an unexpected change in behavior for users, yes?

SkySkimmer:
Added a commit for safety of `coq-prog-args`.
As noted inside I'm not confident about safety for other assistants so I didn't touch them.

I don't know about the default for `proof-prog-name-ask`.

generic/proof-useropts.el
Matafou: Like most security fix :)

SkySkimmer: It's been nil since the 90s, time for a change :v

psteckler: Well, won't it be kind of annoying to be prompted for the program name each time you run PG?

SkySkimmer: In my experience it's OK. It's pre-filled with the default so you just hit RET and go unless you want some other value (in which case there's no way around it or you want better defaults, the later being easier to provide through dir-locals with this change).

Matafou: Isn't it possible to have a configurable set of allowed values, like for File Variables?

SkySkimmer: I removed the proof-prog-name-ask commit from this PR. Continue in #208.

SkySkimmer:
I changed the docstring of `proof-prog-name-ask` from (previous version in this PR)
>*If non-nil, query user which program to run for the inferior process.
>Note that changing this to nil may be a security risk, because a
file might define a malicious ‘proof-prog-name’ as a file local
variable, and you might not notice. Therefore, ‘proof-prog-name’
is considered unsafe if this variable is nil."

to
>*If non-nil, query user which program to run for the inferior process.
>Note that ‘proof-prog-name’ is considered safe if this variable is non-nil.

to make more sense when the default is nil.

Matafou:
If no objection I will mege this today.

coq/coq-system.el
```diff
@@ -39,7 +48,9 @@ This must be a single program name with no arguments; see `coq-prog-args'
 to manually adjust the arguments to the Coq process.
 See also `coq-prog-env' to adjust the environment."
   :type 'string
-  :group 'coq)
+  :group 'coq
+  :safe (lambda (a) (and (stringp a)
+                    (and (boundp 'proof-prog-name-ask) proof-prog-name-ask))))
```
cpitclaudel: `and` can take more than two arguments, so no need to repeat the `and` here; also, there's a macro for this: `bound-and-true-p`

coq/coq-system.el
```diff
@@ -22,6 +22,15 @@
...
   (defvar coq-debug))
 
+;; Arbitrary arguments can already be passed through _CoqProject.
+;; However this is not true for all assistants, so we don't modify the
+;; (defpgcustom prog-args) declaration.
+(defun coq--string-list-p (obj)
+  "Determine if OBJ is a list of strings."
+  (or (null obj) (and (consp obj) (stringp (car obj)) (coq--string-list-p (cdr obj)))))
+
+(put 'coq-prog-args 'safe-local-variable #'coq--string-list-p)
```
cpitclaudel: Is it safe to do this here (are we sure that coq-prog-args is defined at this point?)

easycrypt/easycrypt.el
```diff
@@ -23,7 +23,9 @@
 (defcustom easycrypt-prog-name "easycrypt"
   "*Name of program to run EasyCrypt."
   :type  'file
-  :group 'easycrypt)
+  :group 'easycrypt
+  :safe (lambda (a) (and (stringp a)
+                    (and (boundp 'proof-prog-name-ask) proof-prog-name-ask))))
```
cpitclaudel: Same here

generic/proof-config.el
```diff
@@ -834,7 +834,9 @@ Remark: if `<PA>-prog-args' is non-nil, then `proof-prog-name' is considered
 strictly: it must contain *only* the program name with no option, spaces
 are interpreted literally as part of the program name."
   :type 'string
-  :group 'proof-shell)
+  :group 'proof-shell
+  :safe (lambda (a) (and (stringp a)
+                    (and (boundp 'proof-prog-name-ask) proof-prog-name-ask))))
```
cpitclaudel: And here

generic/proof-useropts.el
```diff
@@ -255,7 +255,9 @@ files which are out of date with respect to the loaded buffers!"
...
-  "*If non-nil, query user which program to run for the inferior process."
+  "*If non-nil, query user which program to run for the inferior process.
+
+Note that ‘proof-prog-name’ is considered safe if this variable is non-nil."
```
cpitclaudel: I think the prefered form of quoting in Emacs is `` `' ``, not `‘’`

cpitclaudel:
I added a few comments.

I don't know how I feel about this change, though. We talked about security: do users of proof-prog-name-ask actually check the command carefully each time? I'd expect not.  I don't, at least — I just press RET out of habit.
Given this, do we expect people to notice a change in that command?

Also: what problem are we trying to solve? IOW, when do we expect people to set proof-prog-name in a file or a `.dir-locals.el`? It makes sense for a compilation command, since that's very file-dependent, but for Coq?

Presumably it'll be a user-specific path, and in that case it's trivial for users to mark specific values as safe.  Am I missing other use cases?

coq/coq-system.el
SkySkimmer: >Is it safe to do this here (are we sure that coq-prog-args is defined at this point?)

There's already `(eval-when-compile (defvar coq-prog-args))` above because it's used in some functions in this file.

Also I don't think it matters whether it's already defined.

coq/coq-system.el
SkySkimmer: >and can take more than two arguments, so no need to repeat the and here; also, there's a macro for this: bound-and-true-p

Sure.

coq/coq-system.el
cpitclaudel: > There's already (eval-when-compile (defvar coq-prog-args)) above because it's used in some functions in this file.

That declares the variable, but it doesn't define it.

> Also I don't think it matters whether it's already defined.

Indeed, good point :) Thanks!

generic/proof-useropts.el
SkySkimmer: I don't know where I even got those quotes from.

SkySkimmer:
>Also: what problem are we trying to solve? IOW, when do we expect people to set proof-prog-name in a file or a .dir-locals.el? It makes sense for a compilation command, since that's very file-dependent, but for Coq?

This PR (and the offshoot #208) was prompted by coq/coq#1136 trying to make the Coq repository's automatic setting of coq-prog-name work with less warnings.
Guessing the right Coq while inside the repository is useful, but maybe dir-locals isn't the right mechanism for all workflows? I don't know.

cpitclaudel:
> Guessing the right Coq while inside the repository is useful, but maybe dir-locals isn't the right mechanism for all workflows? I don't know.

Thanks for pointing to the discussion :) I posted a comment in that thread.

SkySkimmer:
>I don't know how I feel about this change, though. We talked about security: do users of proof-prog-name-ask actually check the command carefully each time? I'd expect not. I don't, at least — I just press RET out of habit.

OK then. That leaves 1 commit with
>Make coq-prog-args safe when list of strings.
>They could be passed through _CoqProject regardless.

SkySkimmer:
Could this PR be merged?

Matafou:
I am a bit lost in the different propositions about the way we can set coq-prog-name and coq-prog-args.
coq/coq#1136 seems cancelled and #6556 seems preferred. So maybe this makes this PR obsolete?

SkySkimmer:
@Matafou it still makes sense to have file local coq-prog-args eg to set `-top` when it matters.

Matafou:
Ok it is orthogonal to other PR. Ok to merge for me.
P

Le ven. 16 févr. 2018 à 12:33, Gaëtan Gilbert <notifications@github.com> a
écrit :

SkySkimmer:
Ping

erikmd:
Thanks @SkySkimmer! this PR will indeed make it easier to set coqtop args as file vars. I've just tested the PR with both `master` and `async` branches and it looks very fine. I'm going to merge it as is.

PR #208, opened 2017-10-24
Default proof-prog-name-ask t

SkySkimmer:
This is probably better for security.

It also aligns with the Emacs compilation-read-command (which has a
similar role).

### DIFF
--- a/generic/proof-useropts.el
+++ b/generic/proof-useropts.el
@@ -254,7 +254,7 @@ files which are out of date with respect to the loaded buffers!"
   :group 'proof-user-options)
 
 (defcustom proof-prog-name-ask
-  nil
+  t
   "*If non-nil, query user which program to run for the inferior process."
   :type 'boolean
   :group 'proof-user-options)


SkySkimmer:
Since this seems controversial it should have its own easy-to-close PR.
Personally I run with this value so it's hard to see why not.

cpitclaudel:
Thanks for the PR!

> Personally I run with this value so it's hard to see why not.

I don't understand :/ Why is it better for security? And why do we want to align with compilation-read-command?

psteckler:
I suppose the security argument is that you know what code you're running, instead of a possibly rogue program in a possibly-altered `PATH`. I wouldn't say that's a big concern.

erikmd:
*Side note:* beyond the security point-of-view, this PR makes me think about the orthogonal problem of switching easily from one Coq version to the other (assuming for instance that we have installed several versions of Coq using `opam`, and we want to process a given theory with several different Coq versions).
To some extend, setting `proof-prog-name-ask` to `t` partly addresses this use case, but maybe we'd want to develop extra facilities to easily switch between different Coq versions? e.g. by automatically displaying a drop-down list extracted from the list of `opam switch`s? What do you think?
(Sorry I'm a bit off-topic with this comment, but if you think it's worth discussing more on that point, maybe we could open an issue for that?)

Matafou:
The security risk is a file containing:
```coq
(*** File Variables: ***)
(*** coq-prog-name "rm" ***)
(*** coq-prog-args '("-rf" "*") ***)
(*** End: ***)
```
But this is already dealt with by File Variable security stuff, so maybe I am over paranoid there.

Matafou:
@erikmd I like the idea.

cpitclaudel:
> The security risk is a file containing […]

I don't really buy this; as you correctly point out, there already is a security mechanism in place for this.
Additionally, I don't think it helps to change the default value: what prevents an attacker from putting `proof-prog-name-ask: nil` in file local variables as well?

erikmd:
@cpitclaudel Thanks for sharing! So I'm going to open an issue to deal with this "switch" feature...

erikmd:
@cpitclaudel said:
> what prevents an attacker from putting proof-prog-name-ask: nil in file local variables as well?

That's true indeed. So changing the default value of `proof-prog-name-ask` to `t` doesn't add much security by itself, beyond the Local Variables security mechanism.

However as I said before, such a confirmation feature when running an external program is "standard" in other modes e.g. `M-x tuareg-run-ocaml` or `M-x compile`. But admittedly, these two examples are a bit different as they correspond to a need for choosing CLI flags to call `ocaml` or `make`...

All things considered, it seems this default value change is not that motivated and we could close this PR for now − while keeping the possibility to reconsider later changing this default value, namely when #210 is integrated and we want to give the user more choices by default when starting Coq in an emacs session :)
So, any objection to close this PR?

Matafou:
No objection to closing.

PR #224, opened 2018-01-24
Look for vernac controls before focus bracket, possible fix for #223

### LINKED ISSUE #223: Time followed by focus bracket gets tokenized wrongly
This issue was reported in coq/coq#6587. In the following script:
```
Goal True.
Time { split. }
Qed.
```
On the second line, PG sends `Time { split.` as a single statement to Coq. It should be two statements, `Time {` followed by `split.`.

This bug affects PG/xml, of course, but also CoqIDE, which has the same fault.

psteckler:
The change here seems to fix the problem, but it should be reviewed.

Fixes #223.

### DIFF
--- a/coq/coq-indent.el
+++ b/coq/coq-indent.el
@@ -378,7 +378,14 @@ Comments are ignored, of course."
       (coq-empty-command-p))
      ;; other bulleting syntax
      ((looking-at "{\\|}\\|-\\|\\+\\|\\*") (coq-empty-command-p))
-     ;; not a bullet, we foound something else, it shoulf be either a
+     ;; vernacular controls Time, Fail, Redirect, Timeout
+     ((or (and (looking-at "e\\>") (looking-back "\\<Tim") (- (point) 3))
+          (and (looking-at "l\\>") (looking-back "\\<Fai") (- (point) 3))
+          (and (looking-at "\"") (looking-back "\\<Redirect\\s-+\"[^\"]+"))
+          (and (looking-at "[[:digit:]]\\_>") (looking-back "\\<Timeout\\s-+[[:digit:]]*")))
+      (goto-char (match-beginning 0))
+      (coq-empty-command-p))
+     ;; not a bullet, we found something else, it should be either a
      ;; dot or the beginning of the file, otherwise we are in the
      ;; middle of a command
      (t (looking-at "\\.\\|\\`"))


psteckler:
Maybe the `Fail` part is silly, does a focus bracket always succeed?

coq/coq-indent.el
```diff
@@ -365,7 +365,7 @@ and return nil."
   "Test if between point and previous command is empty.
 Comments are ignored, of course."
   (let ((end (point))
-        (start (coq-find-not-in-comment-backward "[^[:space:]]")))
+        (start (coq-find-not-in-comment-backward "(Time\\|Fail)?[^[:space:]]")))
```
Matafou: first remark shouldn't those parenthesis be escaped \\(Time\\|Fail\\)?
second remark: I don't see how this could fix things, if we find Time we must recrsively call coq-empty-p to check if there is nothing significant before Time (in which case Time was not the Time command but maybe some coq constant named Time).
The function first looks for something that is significant (non space, non comment) and decides if this is the end of a previous command (which is done recursively). If yes then the command was empty.
`(coq-find-not-in-comment-backward "(Time\\|Fail)?[^[:space:]]")` tries too hard to 

I

psteckler: Sorry, I got this all wrong. I'll try again tomorrow.

psteckler:
I pushed a new commit, which I hope is more sensible.

coq/coq-indent.el
```diff
@@ -378,7 +378,13 @@ Comments are ignored, of course."
       (coq-empty-command-p))
      ;; other bulleting syntax
      ((looking-at "{\\|}\\|-\\|\\+\\|\\*") (coq-empty-command-p))
-     ;; not a bullet, we foound something else, it shoulf be either a
+     ;; Time modifier
+     ((and (> (point) 3)
```
Matafou: LGTM

I would have written it like this
```(and (looking-at "e\\>") (looking-back "\\<Tim")(- (point) 3))```
but nevermind.

psteckler:
There are some more cases to handle, as mentioned in the original issue. I'll add those later today.

psteckler:
@matafou I made changes for the other vernacular controls.

psteckler:
Does someone want to do the merge for this? As the PR author, I don't think I should do it.

Matafou:
Done. Thanks Paul.

Issue #31, opened 2016-01-14
Should we bind C-c C-j in terminals?


cpitclaudel:
`C-c C-RET` (proof-goto-point) doesn't work in shells. Should we bind `C-c C-j` instead on TTYs?

Matafou:
IMHO yes.
P.

2016-01-15 0:14 GMT+01:00 Clément Pit--Claudel <notifications@github.com>:

JasonGross:
I generally bind `C-c RET`, so that `C-c C-RET` works regardless of where I am.

cpitclaudel:
Does your terminal differenciate RET and C-j?

JasonGross:
`C-c C-j` gives "No active hole"

cpitclaudel:
Can you do `C-h k C-c C-j` and `C-h k C-c C-RET`?

JasonGross:
```
C-c C-j runs the command holes-set-point-next-hole-destroy, which is
an interactive compiled Lisp function in `holes.el'.

It is bound to C-c C-j, <menu-bar> <Holes> <Jump To Active Hole>.

(holes-set-point-next-hole-destroy)

Moves the point to current active hole (if any and if in current buffer).
Destroy it and makes the next hole active if any.
```

```
C-c RET runs the command company-coq-proof-goto-point, which is an
interactive compiled Lisp function in `company-coq.el'.

It is bound to C-c RET, C-c <C-return>, <menu-bar> <Proof-General>
<Goto Point>.

(company-coq-proof-goto-point &rest ARGS)

Pass ARGS to `proof-goto-point', hiding company dialog.

[back]
```
(`C-RET` doesn't actually work over mosh at all, but it's easier to teach my fingers to switch back and forth between `C-c C-RET` and `C-c RET` than anything else.)


Using `C-h k C-c C-RET` over ssh in ssh gives `C-c C-^` is not defined.

cpitclaudel:
What a mess.

Matafou:
:) Indeed this is my fault. Holes are my (old) skeleton-like mode. It is
hardly used by anyone other than me I guess, so we can make its shortcuts
different with no harm. And maybe we can remove them completely? It has
more features than company-coq templates but since nobody use them except
me...

Actually I think company-coq should also avoid this shortcut. C-c RET is
something users want to customize (I do: I bind it to dabbrev-expand).

P.






2016-01-18 5:56 GMT+01:00 Clément Pit--Claudel <notifications@github.com>:

[...]

PR #228, opened 2018-02-20
Bind C-c C-m (= C-c RET) to proof-goto-point [tty]

erikmd:
Close ProofGeneral/PG#31

Tested in both GUI and TTY PG (with company-coq)

Can be squashed-and-merged


erikmd:
Thanks for your approval; squash-merging

PR #236, opened 2018-03-01
Fix typos in custom variable descriptions.

tchajed:


erikmd:
Thanks @tchajed for the PR!

erikmd:
@Matafou @cpitclaudel 
The target branch of this PR is `master`
I propose to merge this in `master` once 4e0c2a00 and 61941d93 have been cherry-picked in `master`.
Then we'll be able to cherry-pick 0f84e705 (= this PR) in `async`
Any objection?

cpitclaudel:
Sounds good to me.  Or we could merge master in async.

erikmd:
OK thanks @cpitclaudel, I'll do that a bit later.

> Or we could merge master in async.

Indeed that will be even better :+1:

erikmd:
This makes me realize that if some upcoming PRs propose fixes that could be easily applied in both `master` (proof-shell) and `async` branches, we could maybe suggest to preferably set `master` as target branch, then merge in `async`. Even if one such merge adds a little noise in the history, it is more robust actually.
But naturally, if a change only deals with async PG or would be difficult to port in both branches, sticking to `async` as PR target branch is fine ;)

cpitclaudel:
Yup, absolutely

erikmd:
Done.

PR #340, opened 2018-03-04
fix make html

liyishuai:
`makeinfo` does not support `--split=chapter` option, thus deleted it.

### DIFF
--- a/doc/Makefile.doc
+++ b/doc/Makefile.doc
@@ -18,7 +18,7 @@
 
 MAKE = make -f Makefile.doc 
 MAKEINFO = makeinfo
-TEXI2HTML = makeinfo --html --ifinfo --number-sections --split=chapter --no-headers --css-include=proofgen.css
+TEXI2HTML = makeinfo --html --ifinfo --number-sections --no-headers --css-include=proofgen.css
 # `texinfo-tex' package contains texi2pdf
 TEXI2PDF = texi2pdf
 # `dviutils' package contains these useful utilities.


erikmd:
Hi @liyishuai, thanks for submitting this PR.
But as shown by the small session below in Debian GNU/Linux, my `makeinfo` does support option `--split`. What is your OS/system and version of makeinfo ?

```
$ man makeinfo | grep -A 2 -e "--split="
       --split=SPLIT
              split at SPLIT, where  SPLIT  may  be  `chapter',  `section'  or
              `node'.
$ makeinfo --version
texi2any (GNU texinfo) 6.3

Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

Also, I don't really use `make html` from [this Makefile](https://github.com/ProofGeneral/PG/blob/master/doc/Makefile.doc) in practice, I rather use [that Makefile](https://github.com/ProofGeneral/proofgeneral.github.io/blob/master/doc/Makefile) and if I remove the `--split=chapter` argument from that Makefile, it makes it fail.

BTW if you are interested in some HTML version of the PG documentation, are you aware of that page? <https://proofgeneral.github.io/doc/userman>

liyishuai:
I was using macOS with `makeinfo (GNU texinfo) 4.8`.
I've upgraded to `texi2any (GNU texinfo) 6.5` and now it works.
We should update the [Homebrew formula](https://github.com/Homebrew/homebrew-core/blob/e48dca11cd284bc061d5061a2986a43763da8327/Formula/proof-general.rb#L16) to handle this dependency issue.

erikmd:
@liyishuai
> I was using macOS with makeinfo (GNU texinfo) 4.8.
> I've upgraded to texi2any (GNU texinfo) 6.5 and now it works.

OK! :+1:

> We should update the Homebrew formula to handle this dependency issue.

Are you sure something needs to be updated?
According to [this formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/texinfo.rb#L4) it seems to be already `texinfo 6.5`

Otherwise, feel free to open a PR agains the Homebrew repo :) I don't have a macOS system myself so I couldn't test it.

liyishuai:
Yes, I have created Homebrew/homebrew-core#24971. This should solve the dependency issue.

erikmd:
Thanks @liyishuai! So I can now close #340

PR #342, opened 2018-03-05
Always put newline after conclusion

### LINKED ISSUE #233: Using PG/xml, C-c C-p no longer shows the context for the current goal
@psteckler I have a context and goal that looks like this:

```
1 subgoal

subgoal 1 (ID 37):
  C : Category
  objs : obj_idx → C
  arrmap : M.t (∃ x y : obj_idx, objs x ~{ C }~> objs y)
  dom, i, j : obj_idx
  from, to : ∃ hs : list arr_idx, ReifiedArrow objs arrmap i j hs
  cod : obj_idx
  x : list arr_idx
  p : ReifiedArrow objs arrmap dom cod x
 ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
     ∃ ks : list arr_idx, ReifiedArrow objs arrmap dom cod ks
```

When I use `C-c C-p` now to refresh the display, or after hitting the electric terminator, all that I see in the `*goals*` buffer is:

```
 ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
     ∃ ks : list arr_idx, ReifiedArrow objs arrmap dom cod ks
```

psteckler:
PG/xml was not putting a newline at the end of the \*goals\* buffer when there was a single goal. This confused the regexp search in `coq-goals--show-first-goal`.

Fixes #233 (I think).

### DIFF
--- a/coq/coq-goals.el
+++ b/coq/coq-goals.el
@@ -140,8 +140,7 @@
 	(insert (coq-goals--format-goal-with-hypotheses 
 		 (coq-goals--goal-goal goal1)
 		 (coq-goals--goal-hypotheses goal1)))
-	(when goals-rest
-	  (insert "\n")))
+	(insert "\n"))
       (dolist (goal goals-rest)
 	(cl-incf goal-counter)
 	(insert (format "\nsubgoal %s (ID %s):\n" goal-counter (coq-goals--goal-id goal)))
@@ -170,7 +169,7 @@
 	      (beginning-of-line)
 	      ;; find something backward else than a space: end of conclusion
 	      (ignore-errors (search-backward-regexp "\\S-"))
-	      (recenter (- 1)) ; put bottom of conclusion at bottom of window
+	      (recenter -1) ; put bottom of conclusion at bottom of window
 	      (beginning-of-line)
 	      ;; if the top of conclusion is hidden we may want to show it instead
 	      ;; of bottom of conclusion


erikmd:
@jwiegley could you test whether this PR #342 indeed fixes your reported issue #233 ?

jwiegley:
I will test soon, likely tomorrow.

jwiegley:
It does solve it, but now PG is very "jumpy". I hit `C-c C-RET` to get me to the error (this is actually a trick I use to navigate to the next hole, since it's the quickest way to get to exactly where I need to be), and the screen visible flashes, the window shifts the buffer position, then shifts it back, and then finally quiets down and allows me to edit.  This will get annoying very quickly when doing it hundreds of times a day...

jwiegley:
I guess a principle I'm trying to express is that I'd like PG to continue acting as it does now. If async is enabled, then it's more asynchronous, but it shouldn't feel like a different working environment. The `master` branch still represents the sort of use experience that I want.

erikmd:
> It does solve it, but now PG is very "jumpy".
Hi @jwiegley, which commit did you test ?
Because FYI the commit ecd39d515ef6a92c8202d90e27456994dcbb87b4 of this PR wasn't supposed to fix the behavior of `C-c C-RET` (which has been very improved in #238 AFAICT)

erikmd:
To sum up, do you confirm that the small patch of #342 (proposed by @psteckler just before his leaving) fixes your \*goals\* issue (regarding the hypotheses that were scrolled out of view)?

erikmd:
OK I just saw your other report #346 (thanks for your feedback!) so I assume you tested the current branch async with patch #342 applied; I'll just wait for your confirmation before merging #342 and closing #233 ;)
Erik

jwiegley:
Oh, right, I forget about the goals scrolling! I was looking at whether it resets point to the latest error (which it now does).

jwiegley:
And yes, I'm testing current `master` with #342 applied.

erikmd:
@jwiegley thanks for your reply, but just to be sure:

> I'm testing current `master` with #342 applied.

I guess you mean current `async` with #342 applied, don't you? :) (because the bug #233 solved by this PR didn't impact `master` AFAIK)

Kind regards, Erik

Matafou:
Should we merge?

PR #344, opened 2018-03-07
Add a missing parameter in advice on font-lock-fontify-keywords-region

cpitclaudel:
This breaks various other modes in recent Emacs, most noticeably git-region-history:

Error during redisplay: (jit-lock-function 1501) signaled (wrong-number-of-arguments #[(ad--addoit-function beg end loudly) "\306	\203 \0\307\310\311
!\312\313Q!!\203 \0\310\311
!\312\313Q!
#\210
#\211)\207" [ad-return-value proof-buffer-type proof-assistant-symbol beg end loudly nil fboundp intern symbol-name "-" "font-lock-fontify-syntactically-region" ad--addoit-function] 6] 3)

### DIFF
--- a/generic/proof-syntax.el
+++ b/generic/proof-syntax.el
@@ -236,7 +236,7 @@ this were even more bogus...."
 (eval-after-load "font-lock"
 '(progn
 (defadvice font-lock-fontify-keywords-region
-  (before font-lock-fontify-keywords-advice (beg end loudly))
+  (before font-lock-fontify-keywords-advice (beg end &optional loudly))
   "Call proof assistant specific syntactic region fontify.
 If it's bound, we call <PA>-font-lock-fontify-syntactically-region."
   (when (and proof-buffer-type


erikmd:
Hi @cpitclaudel, thanks for your PR for `master`!
Did you test your patch with `async` as well?

erikmd:
Otherwise I can first merge #238 in `async` (which significantly improves the behavior w.r.t. retracting), and afterwards we test your patch on the new `async` so obtained?

erikmd:
BTW could you precise which recent Emacs version you tested, and which mode contains the function git-region-history ?

cpitclaudel:
I tested the current master, but things have been broken for a while.  The font-lock-fontify-keywords-region has taken the third argument since at least 1995, and our advice was just wrong.
I'm happy with any merge strategy (I don't expect this to conflict)

erikmd:
OK, so thanks for having spotted this longstanding bug!

Just for the record, could you give more details about git-region-history? is it related to Magit?

cpitclaudel:
> Just for the record, could you give more details about git-region-history? is it related to Magit?

No, not at all :) It's called by vc-region-history as vc-git-region-history-mode, which is a function vanilla Emacs.

erikmd:
OK :) I regularly use Emacs' Git facilities (with commands having `C-x v` prefix in vanilla Emacs) but wasn't aware of <kbd>M-x vc-region-history</kbd>, thanks for the tip!

erikmd:
Successfully tested after merging locally this PR in latest `pg: async`. Will merge in both branches.

PR #356, opened 2018-04-17
Fix coq-symbol-face's :inherit attribute

stardiviner:
There is no `default-face` face, it should be `default`. I found this error when I `[M-x org-html-htmlize-generate-css]`.

### DIFF
--- a/coq/coq-db.el
+++ b/coq/coq-db.el
@@ -316,7 +316,7 @@ See `coq-syntax-db' for DB structure."
   :group 'proof-faces)
 
 (defface coq-symbol-face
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Excellent point.  Also, what's the point of inheriting from `default`?

Matafou:
Yes please remove it.

stardiviner:
I updated let it inherit from `font-lock-type-face`.

erikmd:
@stardiviner thanks for your PR!
LGTM; can you please squash your two commits into a single one (and force-push the corresponding branch)?
Then I'll merge it in both `master` and `async`.

PR #375, opened 2018-08-07
Add coq-Print-Ltac to print an Ltac term

jmgrosen:
This is pretty self explanatory; it's just another convenience function that I use pretty often and figure others might too. There's currently no default keybinding.


cpitclaudel:
Thanks!

PR #379, opened 2018-08-17
Support the Variant vernacular

tchajed:
Syntactically looks much like an Inductive, though it is non-recursive
so "where" (mutual recursion) is not supported.


cpitclaudel:
Thanks!

PR #389, opened 2018-09-11
Add 2x splash image for Mac port with Retina display.

mituharu:
Emacs Mac port supports high-resolution (2x) image display for Retina displays.
One way to do that is just to add a file NAME@2x.EXT in parallel with existing NAME.EXT (no need to change Lisp code).
This pull request adds a 2x splash screen image so it looks crispier on Retina displays.

In relation to that, I'd also like to improve toolbar icons on Retina displays (and Dark Mode on macOS 10.14).
README in the `images` directory tells that most of the images are created with Gimp and Inkscape.
If the toolbar icons are originally created as larger bitmaps or resolution-independent data like SVG and the original data are available somewhere, then I can generate high-resolution toolbar icons for Retina displays as multi-image TIFF data that Emacs Mac port prefers.


cpitclaudel:
> If the toolbar icons are originally created as larger bitmaps or resolution-independent data like SVG and the original data are available somewhere, then I can generate high-resolution toolbar icons for Retina displays as multi-image TIFF data that Emacs Mac port prefers.

Slightly off-topic, but: is there any reason not to use SVG icons directly?

mituharu:
> Slightly off-topic, but: is there any reason not to use SVG icons directly?

Because it takes time to setup WebKit (used for rendering SVG images in the Mac port) and makes startup slow.
For this reason, the Mac port uses PNG for the startup Emacs splash image. ("About Emacs" uses SVG, though.)
Also, SVG support in Emacs does not support alpha channel, so it is not suitable for toolbar icons.
I've actually once tried to support alpha channel in the WebKit support code, but I abandoned it because the platform-independent librsvg support code does not do that, and text LCD smoothing requires opaque background.

cpitclaudel:
> Also, SVG support in Emacs does not support alpha channel, so it is not suitable for toolbar icons.

Do you mean just on OSX?  It works fine on GNU/Linux, AFAICT, and I use it in F*-mode.  This is how it looks:

screenshot from 2018-09-11 15-56-15

mituharu:
> > Also, SVG support in Emacs does not support alpha channel, so it is not suitable for toolbar icons.
> 
> Do you mean just on macOS?

Emacs built with GTK+ first tries to let GTK+ render toolbar icon images and falls back on Emacs's own image rendering (see xg_get_image_for_pixmap in gtkutil.c).  That's why SVG in the toolbar on GTK+ supports alpha channel but not inside Emacs buffers.

Matafou:
Do we want to merge this as is? IS there something more to do?
Concerning toolbar icons this could be done in another PR.

Matafou:
This is 5 years old now. Still no real reason not to merge.

Issue #203, opened 2017-09-26
Wrong path send to coqtop through ssh


ildyria:
### TL;DR: On `Require Import` on a project through ssh (tramp) the path sent to coqtop are wrong.

Some informations about my configuration:

Set the `.profile` / `.zshrc` so that Tramp works correctly:

```bash
if [[ $TERM == "dumb" ]]; then	# in emacs
    PS1='%(?..[%?])%!:%~%# '
    # for tramp to not hang, need the following. cf:
    # http://www.emacswiki.org/emacs/TrampMode
    unsetopt zle
    unsetopt prompt_cr
    unsetopt prompt_subst
    unfunction precmd
    unfunction preexec
else
    source $ZSH/oh-my-zsh.sh
fi
```

Set `.spacemacs` parameters:

```emacs-lisp
(defun dotspacemacs/user-config ()
  "Configuration function for user code.
This function is called at the very end of Spacemacs initialization after
layers configuration.
This is the place where most of your configurations should be done. Unless it is
explicitly specified that a variable should be set before a package is loaded,
you should place your code here."
  ;; Open .v files with Proof General's Coq mode
  (load "~/.emacs.d/lisp/PG/generic/proof-site")
  (setq auto-mode-alist (cons'("\\.v$" . coq-mode) auto-mode-alist))
  ;; with tramp this is not required:
  ;; (setq proof-rsh-command "ssh sandor")
  (autoload 'coq-mode "coq" "Major mode for editing Coq vernacular." t)
  ;; coqtop is coqtop (no need for -emacs, it is added by PG)
  (setq coq-prog-name "coqtop")

  (setq tramp-default-method "ssh")
  (with-eval-after-load 'tramp-sh
    (add-to-list 'tramp-remote-path 'tramp-own-remote-path)
    ;; making sure that coqtop is in the path:
    (add-to-list 'tramp-remote-path "/home/benoit/.opam/4.05.0/bin/")
  )
)
```

The following `_CoqProject` will be prepended by `ssh:<server>:<path_to_project>`
```
-Q msl VST.msl
-Q sepcomp VST.sepcomp
-Q veric VST.veric
-Q floyd VST.floyd
-Q progs VST.progs
-Q concurrency VST.concurrency
-Q ccc26x86 VST.ccc26x86
-Q wand_demo wand_demo
-Q sha sha
-Q fcf fcf
-Q hmacfcf hmacfcf
-Q tweetnacl20140427 tweetnacl20140427
[...]

PR #396, opened 2018-10-30
Use non-remote path to expand paths in _CoqProject when file is remote.

dbp:
When editing a remote file, the `coqtop` process will itself be remote, which means that the paths that are passed to it should be _local_, not remote. Otherwise, paths like `/ssh:hostname:/path/to/dir` get passed to `coqtop`, which has no idea what's going on.

I don't know elisp at all, so I'm not sure if there is a better way to write this, but it "seems to work" :) Also, I poked around at the TRAMP docs and couldn't find a more obvious way of doing this, though it seems like it should be a pretty common usecase (as anytime you want to pass arguments to something TRAMP has started, you want to do this renaming).

Also, this relates to (and should close) #203.

### DIFF
--- a/coq/coq-system.el
+++ b/coq/coq-system.el
@@ -547,9 +547,11 @@ variable."
       (let* ((contents (with-current-buffer proj-file-buf (buffer-string)))
              (options (coq--read-options-from-project-file contents))
              (proj-file-name (buffer-file-name proj-file-buf))
-             (proj-file-dir (file-name-directory proj-file-name)))
+             (proj-file-dir (file-name-directory proj-file-name))
+             (proj-file-local-dir (or (file-remote-p proj-file-dir 'localname)
+                                      proj-file-dir)))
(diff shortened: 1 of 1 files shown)


cpitclaudel:
:heart_eyes: thanks!

Yu-zh:
I have got a similar problem here. When I use tramp mode and try to compile an import command, I got an error like this:

`
-*- Mode: compilation; -*-

echo "Require TacticsForTesting." > /var/folders/tf/4md6drms38j4kgv4f3y885k40000gn/T/ProofGeneral-coq1X5zeP.v coqdep -I /ssh:hostname:/path/to/dir/ /var/folders/tf/4md6drms38j4kgv4f3y885k40000gn/T/ProofGeneral-coq1X5zeP.v
Fatal error: exception Sys_error("/ssh:hostname:/path/to/dir/: No such file or directory")
`

How can I fix this? Thanks for your help.

Issue #401, opened 2018-11-12
Mode line disappears


Chobbes:
Let's say I have a file, even just with this:

```
Definition blah {A : Type} (a : A) : list A :=
  cons a nil.
```

Then I hit `C-c C-n` and then `M->` to go to the beginning of the file, and then hit `<C-return>`, then the modeline on the response buffer disappears.

I've also seen the modeline for tho goals buffer randomly disappear as well, and it's pretty weird and jarring when this happens, especially if you resize these buffers with a mouse, or get confused without the mode line separating them.

Executing this is the buffer fixes the problem:

```
(kill-local-variable 'mode-line-format)
```

I think it might have something to do with: https://github.com/ProofGeneral/PG/blob/733cd24a7368ee186884da488da0f59bbedb627e/generic/proof-utils.el#L283

cpitclaudel:
Thanks for opening this issue. I think that code should just go. I don't think hiding the modeline is particularly useful.

Chobbes:
I kind of agree. I'm not really sure why it would ever hide the modeline. I was honestly a little surprised that you *could* hide the modeline :).

Matafou:
This is an old and debatable feature of pg. To save place on small screens I guess. Not fond of it too.

Chobbes:
@Matafou ah okay. I mean, it could always be a toggle, but I think people with small screens (myself included) might want something a bit more predictable, like just deleting the modeline ourselves?

Matafou:
It was more predictable before: there was no modeline at all for the goals and response buffers. At some point it went unpredictable, I don't know why. I think we can remove it.

Chobbes:
Should be fixed now. https://github.com/ProofGeneral/PG/pull/402

PR #402, opened 2018-11-12
Remove code that hides mode-line.

Chobbes:
https://github.com/ProofGeneral/PG/issues/401

This seems to fix the problem! Previously mode-lines would disappear and it was really confusing for me.

### DIFF
--- a/generic/proof-utils.el
+++ b/generic/proof-utils.el
@@ -311,16 +311,7 @@ Ensure that point is visible in window."
 	      ;; inside shrink to fit, for some reason
 	      (when (window-live-p window)
 		(unless (pos-visible-in-window-p (point) window)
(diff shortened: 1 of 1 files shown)


cpitclaudel:
LGTM. Let's leave a few days for people to comment, and then please ping me and I'll merge.

Chobbes:
Sounds good, will do :).

cpitclaudel:
Thanks.

Chobbes:
@cpitclaudel ping as promised!

It's been a few days, and the sentiment seems to be that this would be best removed? Should we wait longer, or do you think it's good to merge?

Thanks! :)

cpitclaudel:
Good for me! Thank :)

PR #410, opened 2018-12-19
Quote ?( ?)

soraros:


Matafou:
Hi, thanks for the PR.

Was the old correct incorrect in some way? It seems to work fine.

Issue #398, opened 2018-11-07
Please improve integration with `use-package`


ejgallego:
When loading proof general using [use-package](https://github.com/jwiegley/use-package) the extra `:no-require t`  has to be passed to the init function:

```elisp
(use-package proof-general
  :no-require t
  :ensure t)
```

otherwise, `use-package` will error: https://github.com/jwiegley/use-package#prevent-a-package-from-loading-at-compile-time

Please either fix or document this requirement for Proof General

erikmd:
Hi @ejgallego, thanks for your report!

We're aware PG's MELPA package still has a few glitches w.r.t. `use-package` support, but we weren't able to polish this yet by lack of time.

BTW I've recently experimented a fully `use-package`-bootstrapped `.emacs` config with PG and `company-coq` in https://github.com/erikmd/coq-init.el (*in French*) which solves the issue you raise by doing a slightly different workaround, using `:defer t` − so I guess your workaround is better than mine.

Anyway this `use-package` config should be simplified after fixing that issue, then we could officially document that PG is `use-package`-compliant in the `README.md` and on https://proofgeneral.github.io/ … I can look at this soon (hopefully next week)

ejgallego:
Thanks @erikmd ! I am not sure which one is the right workaround; I guess you experts will have to tell :)

erikmd:
Indeed that's better; I've updated the repo of my `use-package` config example.

PR #414, opened 2019-01-12
Improve use-package support

erikmd:
This PR amounts to a trivial change (renaming), albeit necessary to simplify PG's setup using `use-package` as pointed out by @ejgallego in ProofGeneral/PG#398.

It also changes the `Commentary` text that will be displayed in http://www.melpa.org/#/proof-general (instead of `No description available.`)

FYI I have prepared a bootstrapped `.emacs` config in https://github.com/erikmd/tapfa-init.el for a course using Tuareg+Merlin+PG+company-coq starting on next Monday, and I'd like to simplify the `use-package` config of PG with this PR before the course starts.

So I plan to merge this on tomorrow Sunday Jan 13 at 12:00 +0100 or so, but if you have comments on the prose, feel free to comment! Cc @ProofGeneral/core

### DIFF
--- a/proof-general.el
+++ b/proof-general.el
@@ -14,10 +14,9 @@
 ;; Package-Requires: ((emacs "24.3"))
 ;; Version: 4.5-git
 
-;; This program is free software; you can redistribute it and/or modify
-;; it under the terms of the GNU General Public License as published by
-;; the Free Software Foundation, either version 3 of the License, or
-;; (at your option) any later version.
+;; This software is free software; you can redistribute it and/or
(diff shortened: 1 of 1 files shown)


monnier:
>   * Rename pg-init.el to proof-general.el

I was about the suggest the same change.

erikmd:
(I've just rebased to fix the merge conflict.)

Matafou:
LGTM

erikmd:
Thanks! so merging now.

PR #418, opened 2019-04-04
Clean -topfile for coq-prog-args properly

SkySkimmer:
This makes it possible to run new and old coq versions consecutively
without manually cleaning coq-prog-args.

It's not the prettiest code but I don't know enough lisp to make it nicer.

### DIFF
--- a/coq/coq-system.el
+++ b/coq/coq-system.el
@@ -378,12 +378,24 @@ options of a few coq-project files does the right thing."
 LOADPATH, CURRENT-DIRECTORY, PRE-V85: see `coq-include-options'."
   (coq-include-options loadpath current-directory pre-v85))
 
+(defun coq--clean-prog-args (args)
+  "Return ARGS without the entries added by `coq-coqtop-prog-args'.
+
+Such entries are currently -emacs and -topfile."
+  (pcase args
+    ((or `("-emacs" . ,rest)
+         `("-topfile" . (,(pred (apply-partially #'equal buffer-file-name)) . ,rest)))
+     (coq--clean-prog-args rest))
+    (`(,car . ,cdr)
+     (cons car (coq--clean-prog-args cdr)))
+    (_ args)))
+
 (defun coq-coqc-prog-args (loadpath &optional current-directory pre-v85)
   "Build a list of options for coqc.
 LOADPATH, CURRENT-DIRECTORY, PRE-V85: see `coq-include-options'."
   ;; coqtop always adds the current directory to the LoadPath, so don't
   ;; include it in the -Q options.
-  (append (remove "-emacs" coq-prog-args)
+  (append (coq--clean-prog-args coq-prog-args)
           (let ((coq-load-path-include-current nil)) ; Not needed in >=8.5beta3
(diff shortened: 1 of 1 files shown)


coq/coq-system.el
```diff
@@ -378,12 +378,26 @@ options of a few coq-project files does the right thing."
 LOADPATH, CURRENT-DIRECTORY, PRE-V85: see `coq-include-options'."
   (coq-include-options loadpath current-directory pre-v85))
 
+(defun coq-clean-prog-args (args)
+  "Return a copy of ARGS with automatically added entries of `coq-prog-args' removed.
+
+Such entries are currently -emacs and -topfile."
+  (if (and (consp args) (equal (car args) "-emacs"))
```
erikmd: The condition `(consp args)` on this line and the subsequent ones could be factored-out, and you could use a `(case ...)`, for example like this:

```lisp
(defun coq-clean-prog-args (args)
  "Return a copy of ARGS with automatically added entries of `coq-prog-args' removed.
Such entries are currently -emacs and -topfile."
  (if (consp args)
      (cond
       ((equal (car args) "-emacs")
	(coq-clean-prog-args (cdr args)))
       ((and (consp (cdr args))
             (equal (car args) "-topfile")
             (equal (cadr args) buffer-file-name))
	(coq-clean-prog-args (cddr args)))
       (t
	(cons (car args) (coq-clean-prog-args (cdr args)))))
      args))
```

coq/coq-system.el
```diff
@@ -378,12 +378,26 @@ options of a few coq-project files does the right thing."
 LOADPATH, CURRENT-DIRECTORY, PRE-V85: see `coq-include-options'."
   (coq-include-options loadpath current-directory pre-v85))
 
+(defun coq-clean-prog-args (args)
+  "Return a copy of ARGS with automatically added entries of `coq-prog-args' removed.
```
erikmd: Thanks for having documented this new function. Two related suggestions:

* Change the first line to
    ```suggestion
  "Return ARGS without the entries added by `coq-coqtop-prog-args'.
    ```
    to cope with usual [guidelines for elisp docstrings](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html):
    > Format the documentation string so that it fits in an Emacs window on an 80-column screen. It is a good idea for most lines to be no wider than 60 characters. The first line should not be wider than 67 characters or it will look bad in the output of apropos. 
    > (…)

* Add a double-hyphen in the identifier of this new function (even this convention is not yet widely used in PG), e.g. `coq--clean-prog-args`.  
    Cf. [this page](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html):
    > Use two hyphens to separate prefix and name if the symbol is not meant to be used by other packages. (…)

coq/coq-system.el
SkySkimmer: I looked around and rewrote using pcase instead

coq/coq-system.el
SkySkimmer: Changed

coq/coq-system.el
erikmd: Thanks! it's even more readable.

PR #419, opened 2019-04-18
Derive proof-mode from prog-mode

bbarenblat:
[prog-mode](https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/progmodes/prog-mode.el?id=a18336a8dc754fa1c68e16dd8009466cf409271b) is the base mode for programming that comes with Emacs: “All major modes for programming languages should derive from this mode so that users can put generic customization on `prog-mode-hook`.”

### DIFF
--- a/generic/proof-script.el
+++ b/generic/proof-script.el
@@ -2305,7 +2305,7 @@ query saves here."
 ;;
 
 ;;;###autoload
-(define-derived-mode proof-mode fundamental-mode
+(define-derived-mode proof-mode prog-mode
   proof-general-name
   "Proof General major mode class for proof scripts.
 \\{proof-mode-map}"


cpitclaudel:
Hmm, I wonder why we hadn't been doing that already.
Thanks for spotting the issue and fixing it!

cpitclaudel:
> Hmm, I wonder why we hadn't been doing that already.

I know now: it's because `prog-mode` is about 12 years younger than `proof-mode`.
 
```
commit 10dcc5612a47d4a6ab9f05c1e8f39993b16bb634
Author: Stefan Monnier <monnier@iro.umontreal.ca>
Date:   Fri May 14 12:41:01 2010 -0400

    * simple.el (prog-mode): New (abstract) major mode.
    * emacs-lisp/lisp-mode.el (emacs-lisp-mode, lisp-mode): Use it.
    * progmodes/sh-script.el (sh-mode): Remove redundant var assignment.
```

PR #426, opened 2019-06-01
Proof diffs revisions requested by @monnier

jfehrle:
Created a hook for coq-insert-tagged-text so that PG code can be loaded even if Coq support code is not loaded.
Updated that routine to process the changes in the buffer rather than from a string.
Updated comments

### DIFF
--- a/coq/coq-diffs.el
+++ b/coq/coq-diffs.el
@@ -29,6 +29,14 @@ For example '<diff.added>foo</diff.added>' inserts 'foo' in the buffer
 and applies the appropriate face.
 
 `coq-tag-map' defines the mapping from tag name to face."
+;; Coq inserts tags before splitting into lines.  Avoid highlighting
+;; white space at the beginning or end of lines in a conclusion or
+;; hypothesis that's split across multiple lines.
+
+;; Doesn't handle the unlikely case of escaping regular text
+;; that looks like a tag.  Unknown tags such as "<foo>" are
+;; shown as-is.  The user can turn off diffs in this very unlikely case.
+
   (let* ((len (length str))
          (off 0)
 	 (fstack)
@@ -38,17 +46,13 @@ and applies the appropriate face.
     (while (< off len)
       (string-match "^\\([ \t]*\\)\\(.*\n?\\)" str off)
       (setq off (match-end 0))
-      ;; FIXME: Why handle whitespace specially?
       (insert (match-string 1 str)) ;; begin-line white space
       (setq rhs (match-string 2 str))
-      ;; FIXME: Why handle whitespace specially?
       (string-match "[ \t\n]*\\'" rhs)
       (let* ((end-white (match-string 0 rhs)) ;; end-line white space
              (line (substring rhs 0 (- (length rhs) (length end-white))))
              (llen (length line))
              (loff 0))
-        ;; FIXME: I don't see any mechanism here to escape text, in case
-        ;; we need to include text that happen to look just like a tag!
         (while (and (< loff llen)
                     (string-match "<\\(/\\)?\\([a-zA-Z\\.]+\\)>" line loff))
           (let* ((end-mark (match-beginning 1))
--- a/coq/coq-diffs.el
+++ b/coq/coq-diffs.el
@@ -15,12 +15,15 @@
 
 ;;; Code:
 
-(defun coq-insert-with-face (str face)
-  (let ((start (point)))
-    (insert str)
-    (if face
-        ;; FIXME: Why not (put-text-property start (point) 'face face)?
-        (overlay-put (span-make start (point)) 'face face))))
+(defun coq-highlight-with-face (end face)
+  (if face
+    ;; (put-text-property (point) end 'face face) doesn't work
+    (overlay-put (span-make (point) end) 'face face))
+  (goto-char end))
+
+(defun coq-search (regex limit)
+  (save-excursion
+     (re-search-forward regex limit t)))
 
 (defun coq-insert-tagged-text (str)
   "Insert text into the current buffer applying faces specified by tags.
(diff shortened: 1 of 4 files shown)


jfehrle:
Attn: @monnier

monnier:
> Created a hook for coq-insert-tagged-text so that PG code can be
> loaded even if Coq support code is not loaded.  Updated that routine
> to process the changes in the buffer rather than from a string.
> Updated comments You can view, comment on, or merge this pull request
> online at:

Thanks Jim,
This looks pretty good, see comments below.

> +;; set hook for diff highlighting routine from non-Coq code
> +(require 'coq-diffs)
> +(setq pg-insert-text-function 'coq-insert-tagged-text)

Can this `setq` be moved to a function somewhere (loading an Elisp file
should ideally not have such side-effects)?  I guess we could put it
inside coq-mode, but maybe better would be to put it near the code that
sends the "Set Diffs" to Coq?

[ Also, I'd recommend you use `#'coq-insert-tagged-text` to clarify the
  intention (you're talking about the function called
  "coq-insert-tagged-text" rather than about the symbol of that name).
  Same for `'insert` further down.
  It makes no difference to the generated code.  ]

> +;; hook for coq diffs highlighting routine
> +(defvar pg-insert-text-function 'insert)

Rather than a comment, you can use a docstring here:

    (defvar pg-insert-text-function #'insert
      "Hook for coq diffs highlighting routine.")

> +(defun coq-search (regex pos)
> +  (let
> +    ((saved-point (point))
> +     (search (re-search-forward regex pos t)))
> +    (goto-char saved-point)
> +    search))  ;; needed?

The indentation looks weird.  Can you reindent it with `C-M-\`?
BTW, You can also write this as

    (save-excursion
      (re-search-forward regex pos t))

And since I'm nitpicking, I'll suggest `limit` instead of `pos` so the
name doesn't just indicate the type of the content but rather the
intended use.

> +  (let* ((fstack)
> +         (lend (make-marker))
> +         (end-white-begin (make-marker)))
> +    (insert str)
> +    (set-mark (point))

Please don't use `set-mark`.  `C-h o set-mark RET` says the following:

    Novice Emacs Lisp programmers often try to use the mark for the wrong
    purposes.  The mark saves a location for the user's convenience.
    Most editing commands should not alter the mark.
    To remember a location for internal use in the Lisp program,
    store it in a Lisp variable.  Example:

       (let ((beg (point))) (forward-line 1) (delete-region beg (point)))."

> +    (backward-char (length str))

I'd have done `(let ((start (point)))` earlier and then `(goto-char
start)` here.

Also, it might be easier to handle the "leading/trailing whitespace"
separately.  E.g. you could first handle the tags and then have a second
pass over the text that does

    (goto-char start)
    (while (re-search-forward "[ \t]*\n[ \t]*" end t)
      (put-text-property (match-beginning 0) (match-end 0) 'face nil))

This way you don't need the loop-nesting because you can ignore
line-boundaries and you can have a single while loop like:

    (goto-char start)
    (setq last start)
    (while (re-search-forward "<\\(/\\)?\\([a-zA-Z\\.]+\\)>" end t)
      (put-text-property last (match-beginning 0) 'face (car fstack))
      (setq last (match-beginning 0))
      (let* ((close-tag (match-beginning 1))
             (tag (match-string 2))
             (face (cdr (assoc tag coq-tag-map))))
        (when face
          (delete-region (match-beginning 0) (match-end 0))
          (setq fstack (if close-tag (cdr fstack) (cons face fstack))))))
    (put-text-property last end 'face (car fstack))

> +        (set-marker end-white-begin (- (- (marker-position lend) end-white-len) nl))

Markers are automatically converted to integers when needed, so you
don't need to explicitly call marker-position here (and in other places
either: it's only used in those rare cases where the code cares about
the difference between a marker and an integer).


        Stefan

jfehrle:
I just pushed a new version.  Are we done yet?

Do you know that Github makes it easy to attach comments to any line in the source code and reply to comments individually?  You do this on the website.  Far easier to work with.  Without this it can get to be quite a chore to respond without getting confused.

I had seen the note about `mark` but that's still quite vague.  Since the response buffer is read-only, does it still use `mark`?

> Can you reindent it with `C-M-\`?

I'm not editing with emacs.  How should it be indented?

> Also, it might be easier to handle the "leading/trailing whitespace" separately.

I'll leave that for someone else to do.  This works fine.  It took me 6 or 8 hours just to change from strings to buffers.  Debugging takes forever.  I will address your other points, though.

monnier:
> I just pushed a new version.  Are we done yet?

Yes.  [ And I didn't mean the previous comments to be requests, just
suggestions.  Sorry that it wasn't clear.  ]

> Do you know that Github makes it easy to attach comments to any line in the
> source code and reply to comments individually?  You do this on the website.

I do know, but that requires running proprietary Javascript code, so
I avoid it as much as I can.
[ I also generally dislike using the browser, to be honest.  Not only
  it's slow, but it's like using someone else's computer: you don't get
  to configure anything at all to your personal preferences.  ]

> Far easier to work with.  Without this it can get to be quite a chore to
> respond without getting confused.

That's pretty how I'd describe using the in-browser interface ;-)

> I had seen the note about `mark` but that's still quite vague.

"The mark" is something that the end user control to set the region and
things like that.  It is implemented as a marker, and Elisp code which
does not specifically want to change the user's mark should use markers
(or just buffer positions when that works as well) rather than using
"the mark".

> Since the response buffer is read-only, does it still use `mark`?

The mark can be used for a variety of purposes, many of which work just
as well in a read-only buffer.

In any case, your new code doesn't touch the mark any more, so it's all good.
I'll point out that instead of using `make-marker` followed by
`set-marker` for `strend` you could also use `copy-marker` which does both
at once.

>> Can you reindent it with `C-M-\`?
> I'm not editing with Emacs.

Sacrilège!

> How should it be indented?

It's very unusual to put the bindings of `let` on their own line,
but for the rare cases where we do it, they're indented by 4, as in:

    (let
        ((x 1)
         (y 2))
      (body))

>> Also, it might be easier to handle the "leading/trailing
>> whitespace" separately.
> I'll leave that for someone else to do.

Fair enough.

Indeed, using overlays you can't just "remove parts" afterwards as
easily as with text-properties (tho I guess you could do that using
`remove-overlays`).

And thanks!

jfehrle:
I changed to use copy-marker and fixed the indentat;ion for the lets.
Ready to merge I think.

> > I'm not editing with Emacs.

> Sacrilège!

I didn't realize you were Mr. Emacs (emeritus).

The last time I used emacs was in 1985-6 when everyone in my group at work
was using it.  It was a reasonable match for the non-bit mapped displays
that were common then. It had more features than vi, but it wasn't easy to
remember all the common keystroke commands!  That has not changed.

It is surprising that it was never adapted for bit-mapped GUIs.  The
version I'm running (Ubuntu in Windows Subsystem for LInux) doesn't have
mouse support enabled out of the box, which I miss a lot.  (Also the
display doesn't always update the display correctly, no doubt a WSL
issue.)  As for elisp, it would help enormously if it told you what line in
the source code has the error.

Regards,

Jim

Matafou:
@jfehrle I don't know what you call a bitmapped gui but emacs (pg actually) looks perfectly bitmapped to me on my desktop:-)

http://cedric.cnam.fr/~courtiep/images/pg.png

I guess the emacs you installed on WSL was the terminal instance only. You can install on windows a graphical version from there https://www.gnu.org/software/emacs/download.html. It works and looks bitmapped at least to me.

jfehrle:
@Matafou The bitmapped version looks good.  I'll keep it in mind for after I get a new laptop--I'm almost out of SSD space and WSL doesn't support external drives.  "Bitmapped GUI" is not entirely redundant--the text-only version of emacs implements GUI concepts such as dropdown menus and multiple buffers.

If @monnier is OK with this version, perhaps one of you fine fellows or gals would approve and merge this cleanup PR?  Thanks.

PR #432, opened 2019-07-25
fix coqtags that can't find some theorem and output empty definition

a0919610611:
As the pull request name
1. The coqtags program can't find some Theorem/Lemma of the following form
```coq
Lemma lemma_name arg1 arg2 : T1 -> T2
```
2. It can't extract the defintion name
```coq
Definition d := true.
``` 
coqtags will output
```
Definition d^?^A1,1
```
The correct one should be
```
Definition d^?d^A1,1
```
I try to fix these two problems.
Thank you for this excellent tool

### DIFF
--- a/coq/coqtags
+++ b/coq/coqtags
@@ -53,7 +53,7 @@ while(<>)
 
 #  print "----- (",$lp,",",$cp,")\n", $stmt, "\n";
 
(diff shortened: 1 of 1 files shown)


Matafou:
LGTM
Thanks!

Scutua (approved):

PR #434, opened 2019-07-26
Backtrack -> BackTo

### LINKED ISSUE #433: Backtrack -> BackTo
Since Coq 8.5 `Backtrack` has been deprecated in favour of `BackTo`.
The latter takes only one argument corresponding to the first one of `Backtrack`.
Even if `Backtrack` still works, the other arguments are ignored, so the transition is trivial and could also lead to some simplification on your side.

Concerned line:  https://github.com/ProofGeneral/PG/blob/6361916d4d80828fd0bcc92c575c59e831b3fa8f/coq/coq.el#L707

CC @maximedenes

gares:
I know nothing about elisp, and I did not test it... please don't laugh ;-)

Fix #433

### DIFF
--- a/coq/coq.el
+++ b/coq/coq.el
@@ -1213,8 +1213,7 @@ be called and no command will be sent to Coq."
      ;; If we go back in the buffer and the number of abort is less than
      ;; the number of nested goals, then Unset Silent and Show the goal
      (and (string-match-p "BackTo\\s-" cmd)
-          (> (length (coq-get-span-proofstack (proof-last-locked-span)))
-             coq--retract-naborts)))
+          (> (length coq-last-but-one-proofstack) coq--retract-naborts)))
     ; "Set Diffs" always re-prints the proof context with (if enabled) diffs
     (list "Unset Silent." (if (coq--post-v810) (coq-diffs) "Show.")))
    ((or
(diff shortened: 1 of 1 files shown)


gares:
See also: https://github.com/coq/coq/pull/10574

erikmd:
Thanks @gares, I'm going to test this with at least two versions of Coq and I let you know.

erikmd:
@gares your fix was fine but actually incomplete (there was another occurrence of "Backtrack", notably in a place where the third argument of "Backtrack m n p." was used in order to figure out how many (nested) goals have been aborted).
So I kept your commit, replaced the other occurrences of Backtrack and also realized there were two bugs in the `coq-empty-action-list-command` function :-)
They are now fixed (see the commit messages for more info).
I've tested the PR with Coq 8.10-beta1 (on a nested-lemmas example + the new "Diffs" menubar of PG) and everything looks fine.
The CI is green but I don't merge yet to give @ProofGeneral/core the opportunity to have a quick look.

gares:
Thanks

erikmd:
On second thought (for bisecting purposes) it was probably better to squash the first 2 commits − done.

Matafou:
Actually I thought backtrack was already removed from pg.
If nested lemmes still work correctly in the last few versions of Coq I strongly support this PR.

erikmd:
Hi @Matafou 
> If nested lemmes still work correctly in the last few versions of Coq

I think so (IINM this feature is deprecated but still maintained, and can be enabled with the option below)

```coq
Set Nested Proofs Allowed.  (* copy-and-paste from one message commit in this PR *)
Lemma foo : True /\ 1 = 1.
  split.
  easy.
  Goal True /\ 2 = 2.
  Proof.
  split; easy.
  Qed.
  easy.
Qed.
```

PR #435, opened 2019-07-31
Correctly run coqidetop

llelf:


erikmd:
@llelf thanks for your PR!
But regarding your first commit, I'm unsure it's necessary to replace `coqtop` with `coqidetop`
(I've just tested the `async` branch with Coq 8.7 and the command-line `/bin/bash -c coqtop -async-proofs on -ideslave -main-channel stdfds 2> /dev/null` does seem to work)

Actually, I thought `coqidetop` was specific to CoqIDE's back-end, but maybe I'm missing something?

@ejgallego @Zimmi48 sorry to bother you, but would you know what is the difference between these two commands?

- `/bin/bash -c coqtop -async-proofs on -ideslave -main-channel stdfds 2> /dev/null`
- `/bin/bash -c coqidetop -async-proofs on -main-channel stdfds 2> /dev/null`

(I did not found some mention of `coqidetop` in the Coq refman)

llelf:
> (I've just tested the async branch with Coq 8.7 and the command-line `/bin/bash -c coqtop -async-proofs on -ideslave -main-channel stdfds 2> /dev/null` does seem to work)

@erikmd You probably meant
`/bin/bash -c 'coqtop -async-proofs on -ideslave -main-channel stdfds' 2> /dev/null` (one argument to `-c`) ;-)

But it no longer works (starting from ≈8.9).

erikmd:
@llelf

> You probably meant […] (one argument to -c) ;-)

Indeed :)

> But it no longer works (starting from ≈8.9)

Good point: I did other tests with the latest point release of Coq 8.8 and 8.9 (8.8.2 and 8.9.1) so to sum up, your patch seems OK for 8.9.1 (I didn't test 8.9.0), while it appears coqidetop is not available for 8.7, 8.8.

So ideally the bugfix should be adapted so that Emacs tests beforehand if Coq >= 8.9, and updates the default value for `coq-prog-name` and the option `-ideslave ` accordingly…
To this aim, one could add a function `coq--post-v89` in `coq-system`, similar to `coq--post-v86`.

Could you update the PR accordingly or do you need more guidance?
Thanks, Erik

erikmd:
Actually I guess the fix won't be as simple as this, because the autodetection of the coq version itself depends on `coq-prog-name`… so to avoid this vicious circle, maybe we should replace
https://github.com/ProofGeneral/PG/blob/d7d6035825671fc714ddf8f044ad7df3239e7dab/coq/coq-system.el#L198
and check the version from "coqc -v" instead ?
(actually, an optimized version of this algorithm could be: check the version from "coqc -v" if `coq-prog-name` has not be customized, otherwise check the version from `coq-prog-name`)

erikmd:
Anyway, I am afraid there are still several bugs in the `async` branch currently that prevent it from being usable enough… out of curiosity, @llelf have you tried both `master` and `async` branches of PG or did you directly start using `async`?

In any case, thank you very much for having reported this coqidetop-related issue :+1: we weren't aware of that.

llelf:
> (actually, an optimized version of this algorithm could be: check the version from "coqc -v" if coq-prog-name has not be customized, otherwise check the version from coq-prog-name)

I will look at it.

> out of curiosity, @llelf have you tried both master and async branches of PG or did you directly start using async?

I was just curious about XML/SerAPI things.  `async` is buggy is so-almost-usable. 
Normally I use (well…starting to use, I’m only at the beginning of my coq journey) PG/master.

Zimmi48:
Your analysis is right coqidetop needs to be used for any IDE support through the XML protocol starting with Coq 8.9, not just for CoqIDE. See example of users: https://github.com/EugeneLoy/coq_jupyter/ and https://github.com/whonore/Coqtail.

https://github.com/EugeneLoy/coq_jupyter/blob/dd768408ee702fc180a21c07a589b3e160450ba5/coq_jupyter/coqtop.py#L48-L63

https://github.com/whonore/Coqtail/blob/71c944a71675fcc9a19e805e469eef62544c8c30/python/xmlInterface.py#L1261-L1263

llelf:
@erikmd something like this?

I had to move some code (that's https://github.com/ProofGeneral/PG/pull/435/commits/9b39f3e877f2ab34e970ecb08993ffe90848c484) + the actual change is https://github.com/ProofGeneral/PG/compare/async...llelf:318cecc8

> async is buggy is so-almost-usable.

I meant “async is buggy **but** so-almost-usable.”

erikmd:
Hi @llelf,
thanks for your update! and sorry for the delay… this beginning of semester is very busy, but I plan to review your PR soon, hopefully next week-end (otherwise feel free to ping me)

erikmd:
Dear @llelf , sorry I had lost track of this PR, but I've just made several experiments and commits in your branch, and the PR looks in a very good shape now, so let's merge it!

PR #443, opened 2019-10-07
Make -topfile facility work for unnamed files

pi8027:
Currently, `-topfile` support of Coq PG doesn't work if `buffer-file-name` is `nil`, e.g., when proving in the `*scratch*` buffer. This PR changes it to do not use `-topfile` argument if `buffer-file-name` is `nil`.

### DIFF
--- a/coq/coq-system.el
+++ b/coq/coq-system.el
@@ -432,7 +432,8 @@ LOADPATH, CURRENT-DIRECTORY, PRE-V85: see `coq-include-options'."
   "Build a list of options for coqc.
 LOADPATH, CURRENT-DIRECTORY, PRE-V85: see `coq-coqc-prog-args'."
   (append
-   (if (coq--supports-topfile) (cons "-topfile" (cons buffer-file-name nil)) "")
+   (if (and (coq--supports-topfile) buffer-file-name)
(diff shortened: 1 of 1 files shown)


erikmd:
Thanks @pi8027! your patch LGTM.
Even if this might be tagged as a feature PR (to be able to prove in the `*scratch*` buffer as you say :),
I guess we should rather put the `fix` label here.

coq/coq-system.el
```diff
@@ -432,7 +432,8 @@ LOADPATH, CURRENT-DIRECTORY, PRE-V85: see `coq-include-options'."
   "Build a list of options for coqc.
 LOADPATH, CURRENT-DIRECTORY, PRE-V85: see `coq-coqc-prog-args'."
   (append
-   (if (coq--supports-topfile) (cons "-topfile" (cons buffer-file-name nil)) "")
+   (if (and (coq--supports-topfile) buffer-file-name)
+       (cons "-topfile" (cons buffer-file-name nil)) "")
```
cpitclaudel: Btw (not from this pr) this cons cons thing can be written `(list "-topfile" buffer-file-name)`

PR #447, opened 2019-12-08
Recognize "Timeout" before save keywords to fix #446

### LINKED ISSUE #446: Timeout Defined is misclassified resulting in strange parsing errors after backtracking into the middle of a proof
```
$ cat foo.v
(* -*- coq-prog-args: ("-q") -*- *)
Goal True = True.
  idtac.
  Timeout 1 reflexivity.
  Timeout 1 Defined.
BackTo 3.
Timeout 1 reflexivity.
(*
Error:
Syntax error: [located_vernac] expected after [prim:natural] (in [vernac:Vernac.vernac_control]).
*)
$ coqtop -q -emacs < foo.v
Welcome to Coq 8.10.0 (October 2019)
Skipping rcfile loading.

<prompt>Coq < 1 || 0 < </prompt>1 subgoal (ID 3)

  ============================
  True = True

<prompt>Unnamed_thm < 2 |branch| 0 < </prompt>
<prompt>Unnamed_thm < 3 |branch| 0 < </prompt>No more subgoals.

<prompt>Unnamed_thm < 4 |branch| 0 < </prompt><infomsg>Unnamed_thm is defined</infomsg>

<prompt>Coq < 5 || 0 < </prompt>1 subgoal (ID 3)

  ============================
  True = True

<prompt>Unnamed_thm < 6 || 0 < </prompt>Toplevel input, characters 10-21:
> Timeout 1 reflexivity.
>           ^^^^^^^^^^^
Error:
Syntax error: [located_vernac] expected after [prim:natural] (in [vernac:Vernac.vernac_control]).

<prompt>Unnamed_thm < 6 || 0 < </prompt>
```

Note that this situation can occur in interactive use when I end a proof with `Timeout Defined`.  By contrast, if I end a proof with `Time Defined` or `Defined`, then trying to retract the buffer automatically jumps to before the beginning of the proof.  I'm assuming this is a PG bug because the error from coqtop occurs whenever I use `BackTo` to go to the middle of the proof, and it seems that PG just issues different commands depending on how I end the proof.

erikmd:
This PR should fix #446 (tested with Emacs 25.1.1 on the MWE given by @JasonGross)

Cc @ProofGeneral/core can one of you approve the patch?

then we can merge it no later than on Monday.

Thanks again Jason for the report!

### DIFF
--- a/coq/coq-syntax.el
+++ b/coq/coq-syntax.el
@@ -1070,7 +1070,7 @@ It is used:
 ;; It is understood here as being a goal.  This is important for
 ;; recognizing global identifiers, see coq-global-p.
 (defconst coq-save-command-regexp-strict
-  (concat "\\`\\(?:Time\\s-+\\)?\\("
+  (concat "\\`\\(?:Time\\s-+\\|Timeout\\s-+[[:digit:]]+\\s-+\\)*\\("
           "\\_<"
 	  (regexp-opt coq-keywords-save-strict)
 	  "\\_>"
@@ -1293,12 +1293,13 @@ It is used:
 
 (defconst coq-save-command-regexp
   ;; FIXME: The surrounding grouping parens are probably not needed.
-  (concat "\\`\\(\\(?:Time\\s-+\\)?\\_<"
+  (concat "\\`\\(\\(?:Time\\s-+\\|Timeout\\s-+[[:digit:]]+\\s-+\\)*\\_<"
            (regexp-opt coq-keywords-save t)
            "\\_>\\)"))
 
 (defconst coq-save-with-hole-regexp
-  (concat "\\(?:Time\\s-+\\)?\\(" (regexp-opt coq-keywords-save-strict)
+  (concat "\\(?:Time\\s-+\\|Timeout\\s-+[[:digit:]]+\\s-+\\)*\\("
+          (regexp-opt coq-keywords-save-strict)
 	  "\\)\\s-+\\(" coq-id "\\)\\s-*\\."))
 
 (defconst coq-goal-command-regexp


JasonGross:
Thanks!  This isn't quite enough though.  For example, there's a similar problem with `Time Time Defined`, and I bet there's also the same problem with `Time Timeout 1 Defined` and `Timeout 1 Time Defined`.  (There is also a problem with `Time (* foo *) Defined`, but I imagine that one might be harder to fix, and it seems fine to declare comments in that location basically unsupported, perhaps.)

coq/coq-syntax.el
```diff
@@ -1070,7 +1070,7 @@ It is used:
 ;; It is understood here as being a goal.  This is important for
 ;; recognizing global identifiers, see coq-global-p.
 (defconst coq-save-command-regexp-strict
-  (concat "\\`\\(?:Time\\s-+\\)?\\("
+  (concat "\\`\\(?:Time\\s-+\\|Timeout\\s-+\\(?:[[:digit:]]+\\s-+\\)?\\)?\\("
```
JasonGross:
```suggestion
  (concat "\\`\\(?:Time\\s-+\\|Timeout\\s-+[[:digit:]]+\\s-+\\)?\\("
```
The number is mandatory after `Timeout`.  (What's up with `\\s-+`, though?  what's the `-`?)

coq/coq-syntax.el
```diff
@@ -1293,12 +1293,13 @@ It is used:
 
 (defconst coq-save-command-regexp
   ;; FIXME: The surrounding grouping parens are probably not needed.
-  (concat "\\`\\(\\(?:Time\\s-+\\)?\\_<"
+  (concat "\\`\\(\\(?:Time\\s-+\\|Timeout\\s-+\\(?:[[:digit:]]+\\s-+\\)?\\)?\\_<"
```
JasonGross:
```suggestion
  (concat "\\`\\(\\(?:Time\\s-+\\|Timeout\\s-+[[:digit:]]+\\s-+\\)?\\_<"
```

coq/coq-syntax.el
```diff
@@ -1293,12 +1293,13 @@ It is used:
...
            (regexp-opt coq-keywords-save t)
            "\\_>\\)"))
 
 (defconst coq-save-with-hole-regexp
-  (concat "\\(?:Time\\s-+\\)?\\(" (regexp-opt coq-keywords-save-strict)
+  (concat "\\(?:Time\\s-+\\|Timeout\\s-+\\(?:[[:digit:]]+\\s-+\\)?\\)?\\("
```
JasonGross:
```suggestion
  (concat "\\(?:Time\\s-+\\|Timeout\\s-+[[:digit:]]+\\s-+\\)?\\("
```

coq/coq-syntax.el
erikmd: > The number is mandatory after `Timeout`

good catch, thanks

>  what's the `-`?

this is specific to GNU Emacs regexps:

https://www.gnu.org/software/emacs/manual/html_node/emacs/Regexp-Backslash.html#Regexp-Backslash

> **`\sc`**  
>     matches any character whose syntax is **`c`**. Here **`c`** is a character that designates a particular syntax class: thus, ‘w’ for word constituent, ‘-’ or ‘ ’ for whitespace, ‘.’ for ordinary punctuation, etc.

erikmd:
@JasonGross 

> there's a similar problem with `Time Time Defined`, and I bet there's also the same problem with `Time Timeout 1 Defined` and `Timeout 1 Time Defined`

You're right that supporting `Time Timeout 1 Defined` and `Timeout 1 Time Defined` is very important!

But would it be useful to support `Time Time Defined`? and `Time Time Time Defined`? :)

Anyway, even if it is not useful to cover all these cases, for the sake of simplicity of the regexp, I can just as well put a `*` in place of the `?`. Would this choice look good to you?

erikmd:
Hi @ProofGeneral/core, unless someone objects to the current fix, I propose to merge this PR tonight.

Matafou:
No objection. Thanks!

cpitclaudel:
Same :+1:

erikmd:
OK thanks! merging now.

PR #451, opened 2020-01-08
Add do notation support for better automatic indentation.

Chobbes:
There's lots of Coq projects that really rely upon having haskell-like do-notation (e.g., anything that uses extlib monads). The current proof-general makes this pretty awkward because it doesn't handle indenting these correctly. For instance it might do this:

```
      match n with
      | O => ret y
      | S n =>
        v <- blah;;
          v' <- foo ;;
          ret v'
      end.
```

instead of this:

```
      match n with
      | O => ret y
      | S n =>
        v <- blah;;
        v' <- foo ;;
        ret v'
      end.
```

This pull request addresses this issue. I'm not 100% sure if the precedence is exactly right, but I'm happy to make any changes. This seems to work well, though.

I have never been happier about a small change to something --- this makes working in Coq so much nicer :).

### DIFF
--- a/coq/coq-smie.el
+++ b/coq/coq-smie.el
@@ -51,19 +51,19 @@
 ;     ,@body
 ;     (message "%.06f" (float-time (time-since time)))))
 
-(defcustom coq-smie-user-tokens nil
+(defcustom coq-smie-user-tokens '(("<-" . "monadic bind") (";;" . "monadic sequence"))
   "Alist of (syntax . token) pairs to extend the coq smie parser.
-These are user configurable additional syntax for smie tokens.  It
-allows to define alternative syntax for smie token.  Typical
+These are user configurable additional syntax for smie tokens.
+It allows to define alternative syntax for smie token.  Typical
 example: if you define a infix operator \"xor\" you may want to
 define it as a new syntax for token \"or\" in order to have the
 indentation rules of or applied to xor.  Other exemple: if you
 want to define a new notation \"ifb\" ... \"then\" \"else\" then
 you need to declare \"ifb\" as a new syntax for \"if\" to make
-indentation work well."
+indentation work well. The default list includes extlib style
+monad notations."
   :type '(alist :key-type string :value-type string)
   :group 'coq)
-
 
 (defalias 'coq--string-suffix-p
   ;; Replacement for emacs < 24.4, borrowed from sindikat at
@@ -874,9 +874,11 @@ Typical values are 2 or 4."
 ;;;
        ("(" exp ")") ("{|" exps "|}") ("{" exps "}")
        (exp "; tactic" exp) (exp "in tactic" exp) (exp "as" exp)
+       (exp "monadic sequence" exp)  ;; Stand in for monadic sequencing notation.
        (exp "by" exp) (exp "with" exp) (exp "|-" exp)
        (exp ":" exp) (exp ":<" exp) (exp "," exp)
        (exp "->" exp) (exp "<->" exp) (exp "&" exp)
+       (exp "monadic bind" exp)  ;; Stand in for monadic binding notation.
        (exp "/\\" exp) (exp "\\/" exp)
        (exp "==" exp) (exp "=" exp) (exp "<>" exp) (exp "<=" exp)
        (exp "<" exp) (exp ">=" exp) (exp ">" exp)
@@ -952,7 +954,9 @@ Typical values are 2 or 4."
       (assoc "; tactic") (assoc "in tactic") (assoc "as" "by") (assoc "with")
       (assoc "|-") (assoc ":" ":<") (assoc ",")
       (assoc "else")
+      (assoc "monadic sequence")
       (assoc "->") (assoc "<->")
+      (assoc "monadic bind")
       (assoc "&") (assoc "/\\") (assoc "\\/")
       (assoc "==") (assoc "=") (assoc "<" ">" "<=" ">=" "<>")
       (assoc "=?") (assoc "<=?") (assoc "<?") (assoc "^")


vzaliva:
this patch would be exta-handy!

Matafou:
Here you go, just put this in your .emacs:

(setq coq-smie-user-tokens '(("<-" . "&")(";;" . "->")))

What do you think? Is it better than hard-wiring these notation?
If you think this notation is standard enough to be hard-wired I merge this PR right away.

By the way I just noticed that /\ and \/ are wrongly prioritized. Strange nobody complained. Maybe very few people use indentation after all??

cpitclaudel:
> By the way I just noticed that /\ and / are wrongly prioritized. Strange nobody complained. Maybe very few people use indentation after all??

I think most people do, but `/` isn't that common on its own, is it?

> (setq coq-smie-user-tokens '(("<-" . "&")(";;" . "->")))

This is neat; we should document it in the readme, I think.

Matafou:
Sorry I meant \\/ and /\

cpitclaudel:
> Sorry I meant \/ and /\

Ah, I always parenthesize them for clarity

YaZko:
> > (setq coq-smie-user-tokens '(("<-" . "&")(";;" . "->")))
> 
> This is neat; we should document it in the readme, I think.

Indeed that sounds like a satisfying solution as well! Some documentation would definitely be very welcome to help us craft such extensions to the indenting mechanism ourselves, I am not sure how I should read these tokens.

Thanks for the trick!

vzaliva:
I’ve noticed one problem with proposed indentation patch:

```
Definition foo (a:option nat) :=
  if true then ret tt
  else a ;;
  ret (tt).
```

This is not the right indentation. It looks like (if ... then ... else ...) ;; ret. Adding parentheses fixes the indentation:

```
Definition foo' (a:option nat) :=
  if true then ret tt
  else (a ;;
        ret (tt)).
```

So it looks like additional fixes to grammar are needed. 

But my vote is to include this notation in standard PG Coq grammar as it's quite common.

YaZko:
It would definitely be comfy to have this notation handled by default. 

However orthogonally to it, having a generic documented way to extend indenting support to user-defined notations would be wonderful.

Matafou:
> I’ve noticed one problem with proposed indentation patch:
> 
> ```
> Definition foo (a:option nat) :=
>   if true then ret tt
>   else a ;;
>   ret (tt).
> ```

Did you remove your patch first? I don't have this behaviour.

Matafou:
> Indeed that sounds like a satisfying solution as well! Some documentation would definitely be very welcome to help us craft such extensions to the indenting mechanism ourselves, I am not sure how I should read these tokens.

It allows to tell the indentation mechanism: "treat this token as this other one". I.e. same associativity, same priority.  That is the meaning of each pair: the first component is a new token, the second is the already defined one.

It is thus quite limited: you can only define a "synonym" (indetation-wise) of another token.

We could enhance this mechanism by defining lots of dummy tokens (with no syntax by default), just there to be "plugable".

Matafou:
> > I’ve noticed one problem with proposed indentation patch:
> > ```
> > Definition foo (a:option nat) :=
> >   if true then ret tt
> >   else a ;;
> >   ret (tt).
> > ```

Wait. Is this ";;" legal here anyway? There is no "<-".

And what should be the indentation of:
```
Definition foo' (a:option nat) :=
    x <- if true then ret tt
         else a ;;
    ret (tt). (*??*)
```

Matafou:
I have the feeling that we could define at least 2 sets of generic monadic tokens with different priorities to allow people to plug to one or another.

Probably we also need another plugable token for the other usual monadic notation with a prefix keword:  "do x <- e ; ...".

Matafou:
I just started a poll on discord about monadic notations (and @YaZko already answered:-). We will see if there is a consensus. I am always unconfortable with non builtin syntax in coq made builtin in pg: it always end up making someone angry (using "<-" for something else for instance).

https://coq.discourse.group/t/poll-which-monadic-notation-do-you-use/557?u=matafou

vzaliva:
> Wait. Is this ";;" legal here anyway? There is no "<-".

Sorry, I've over-simplified my example. But this does not change the indetnation:

```
Definition foo (a:option nat) :=
  if true then ret tt
  else x <- a ;;
  ret (tt).

Definition foo' (a:option nat) :=
  if true then ret tt
  else (x <- a ;;
        ret (tt)).
```

But both these functions are identical (proof by reflexivity) and should be indented the same way.

> And what should be the indentation of:
> 
> ```
> Definition foo' (a:option nat) :=
>     x <- if true then ret tt
>          else a ;;
>     ret (tt). (*??*)
> ```

The function above does not parse using Ext-Lib monadic notations:

```
Error: Syntax error: ';;' expected after [constr:operconstr] (in [constr:operconstr]).
```

So preferred indentation is _undefined_ :)

YaZko:
> I just started a poll on discord about monadic notations (and @YaZko already answered:-). We will see if there is a consensus. I am always unconfortable with non builtin syntax in coq made builtin in pg: it always end up making someone angry (using "<-" for something else for instance).
> 
> https://coq.discourse.group/t/poll-which-monadic-notation-do-you-use/557?u=matafou

I don't want to shoot my own faction in the foot, but I'd tend to agree that having a mechanism to do the extension on the user side probably makes more sense. 
I'll toy around with your trick  of declaring synonymous tokens, that seems a bit hacky but promising :)

Thanks also for setting up the poll!

Matafou:
I think a good solution is a generic mechanism WITH a default value to ext-lib syntax.
Basically the generic mechanism is: have dummy tokens (e.g. "monadic op1" "monadic op2" etc) with no real syntax attached to them in the smie grammar. And then the user-token stuff defines "<-" ans ";;" as syntax for them. Having default user-token is OK, people using these symbols for sometinhg else will just have to define an empty user-token.

Matafou:
@Chobbes what do you think? Do you want to write your PR according to this idea? Do you want me to do it?
But first I need to understand exactly what should be the precedence of <- and ;;. In particular I am not convinced by your example:
```
Definition foo (a:option nat) :=
  if true then a
  else b;;
  ret (tt).
```
could equally mean both: 
```
Definition foo (a:option nat) :=
  (if true then a
  else b);;
  ret (tt).
```
and
```
Definition foo (a:option nat) :=
  if true then a
  else (b;;
        ret (tt)).
```
why should the latter be preferred?

Chobbes:
Hmmm. I'm not really sure which should be preferred either -- both seem reasonable :|.

I can update this to have the generic monadic op tokens. Feedback is of course welcome as I'm a bit new to this smie infrastructure, but I would really like to learn a bit more about proof general internals for the future :). Give me a bit. I'll try to make these changes now.

vzaliva:
I think it is not a matter of preference but a matter of semantics. If we
take ExtLib's implementation of monad notation as the reference it clearly
the 2nd.

Bracketed "else" is idented as expected and removing parentheses does not
change the semantics so I would expect the indentation to stay  the same.

Chobbes:
@Matafou I messed with this a little bit, but I wasn't sure how to get notation with "do" to work properly. I.e., you would want the following to indent something like:

```
do a <- ma;;
     b <- mb;;
     ret (a, b)
```

Or even:

```
do
  a <- ma;;
  b <- mb;;
  ret (a, b)
```

```
Definition blah := do
     a <- ma;;
     b <- mb;;
     ret (a, b)
```

But, I couldn't figure out how to make this work... I tried adding a "monadic start" token for the do part, and then added rules for this, but this broke indentation for the rest of the notation kind of like how it's currently broken...

The other problem is the notations that seem to be floating around that use `;` --- this causes issues with the `; tactical` rules.

I changed the precedence so that if / else indents (I believe) as expected, @vzaliva is totally right that this should match what the notations actually do.

Any thoughts?

Matafou:
Of course we follow ext-lib notation so we should also follow its precedence.

In Compcert there is no grouping of "<-", there is one "do" per "<-". May be this will simplify things?

I will have a look at the ";" thing. If it interact really badly with ";" of tactics we may need to do hackier things (i.e. hack the lexer)...

Beware: coq grammar being a nightmare indentation code of pg is thus awful even using the great smie stuff. It is almost clean if you don't look at the lexer (which sadly is 80% of indentation code :-().

Chobbes:
One "do" per "<-" is probably better.

The main thing with ";" is just that if you have...

```
tactic;
  other;
  tactics;
```

It will instead do:

```
tactic;
other;
tactics;
```

At least, that's the interaction that I know about!

Matafou:
Indeed reusing the ";" token breaks the tactical indentation. I need to first redefine the detection of the tactical ";" token in the lexer.
Can you remove the ";" stuff? I will try to add it later.
P.

Matafou:
Please have a look at PR #452, which should subsume this one.
Both syntaxes seem to work, and I detected no breakage of other indentation features.

Matafou:
Still having a small glitch in the lexer with the "<-" token from "rewrite <-". It does not seem to impact indentation very often but anyway...

Matafou:
May I close this now? This PR inspired mine which has been merged (and time will test it).

Chobbes:
Sounds good to me!

Calvin Beck

Matafou:
Thanks for the suggestions and PR.

Chobbes:
Thanks for running with it and making it better :).

PR #452, opened 2020-01-19
Generic monadic indentation + specifically ext-lib / Compcert + doc.

Matafou:
This a generalization of PR#451 proposed by @Chobbes.

Since these syntax are not completely universal (and not builtin in
coq), the idea is to make the syntax customizable, especially to make
it possible to disable it.

NOTE: to make the Compcert syntax supported I had to refine the smie
lexer so that the ";" token was emitted as a fllback instead of ";
tactic".

NOTE2: I had to make the coq-user-token and coq-monadic-tokens be
tested ON THE RESULT of smie-coq-backward-token.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -60,6 +60,22 @@ and the PG Trac http://proofgeneral.inf.ed.ac.uk/trac
     
     The set of tokens can be seen in variable smie-grammar.
 
+*** Indentation of monadic notations.
+    Using the extensibility for indentation described above we provide
+    a way to define your own monadic operators using the
+    coq-smie-monadic-tokens in the same spirit as coq-smie-user-tokens
+    above.
+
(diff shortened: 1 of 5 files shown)


Matafou:
This is an alternative proposition to @chobbes 's PR #451 to also support compcert-like monadic syntax. It needed to refine the lexer.

Chobbes:
This looks great! Thank you for running with this :).

Matafou:
Merging this, it does not seem to break indentation too much...
More indentation fixes coming, following my [poll](https://coq.discourse.group/t/poll-which-monadic-notation-do-you-use/557?u=matafou) on discourse.

vzaliva:
Thanks!

Scutua (approved):

PR #468, opened 2020-03-12
measure difference of silent and non-silent processing

hendriktews:
These are just two lines for measuring how long it takes to assert a block. The numbers show up in the *Messages* buffer. To see the difference between silent and non-silent processing, do it twice with and without proof-full-annotation set (also available in menu PG->Quick Options->Processing->Full Annotation).
I see 10.7 seconds for silent and 10.9 seconds for ≈3000 lines, without loading required modules.

### DIFF
--- a/generic/proof-shell.el
+++ b/generic/proof-shell.el
@@ -1026,6 +1026,7 @@ being processed."
 	  (nconc proof-action-list queueitems))
     
     (when nothingthere ; process comments immediately
+      (message "START PG QUEUE %s" (format-time-string "%H:%M:%S.%1N"))
       (let ((cbitems  (proof-shell-slurp-comments)))
 	(mapc 'proof-shell-invoke-callback cbitems)))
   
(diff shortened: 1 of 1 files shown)


erikmd:
Hi @hendriktews ! just a minor remark regarding this new feature of GitHub PRs (but maybe you're already aware of it): we can create − or convert, see screenshot below − PRs in "draft mode" to automatically tag them as not-to-be-merged-yet.

2020-05-28_19-57-28_Screenshot_PR-draft-mode

erikmd:
@hendriktews just FYI I converted this PR in "draft mode" as I had proposed in my previous comment

hendriktews:
Erik Martin-Dorel <notifications@github.com> writes:

> @hendriktews just FYI I converted this PR in "draft mode" as I had proposed in my previous comment

The discussion in PR #467 has concluded some time ago already
that silent mode is preferred. Therefore we should abandon this
PR.

erikmd:
> The discussion in PR #467 has concluded some time ago already
that silent mode is preferred. Therefore we should abandon this
PR.

OK so closing that PR, thanks @hendriktews

Issue #477, opened 2020-04-02
Dune support


ejgallego:
Hi folks,

the Coq mode for Dune is getting more mature, and a big blocker now is integration with Proof General. I am unsure how to best approach this; basically PG should add the right flags to make libraries compiled in `_build` into scope.

A possibility is indeed to let `dune` handle this; so PG could call `dune coqtop foo.v` and indeed coqtop would be started with the right settings. WDYT?

cpitclaudel:
> A possibility is indeed to let dune handle this; so PG could call dune coqtop foo.v and indeed coqtop would be started with the right settings. WDYT?

Sounds reasonable, but we'd need to be able to determine when dune is in use.

Could dune generate a _CoqProject file for us? In my projects I just write it by hand:

```
-R _build/default/coq/ Xyz
-Q _build/default/tests/ Tests
-Q _build/default/examples/ Examples
-R _build/default/examples/rv RV
```

Also, we need some indication of where each .v file came from, so that jump-to-definition can go to the right place.

ejgallego:
> Could dune generate a _CoqProject file for us?

It could, but I'm not sure I'm willing to do this; notably the `_CoqProject` format will have trouble keeping with Dune features; also it pollutes the source tree. Experience with generating `.merlin` files in OCaml indicates that this is a path we don't want to follow.

> Also, we need some indication of where each .v file came from, so that jump-to-definition can go to the right place.

What do you concretely mean? If you use `dune coqtop foo.v` you would be using the right file already; otherwise you could implement something like tuareg does [it understands _build]

`dune` provides some APIs to query projects, but I guess I'd need to see the concrete use case. I'm not sure if this index functionality is something to have in PG, or instead PG should rely on an external too such as `ocp-index` for OCaml.

ejgallego:
> notably the `_CoqProject` format will have trouble keeping with Dune features

Yeah, already we cannot generate a `_CoqProject` that works for a simple case with two composed libraries, and those with different flags.

cpitclaudel:
> It could, but I'm not sure I'm willing to do this; notably the _CoqProject format will have trouble keeping with Dune features; also it pollutes the source tree. Experience with generating .merlin files in OCaml indicates that this is a path we don't want to follow.

OK. Then running `dune coq` sounds good. How do we detect dune projects reliably? Do we just search for `dune-project` files?

> What do you concretely mean?

[...]

PR #478, opened 2020-04-03
coq-par-compile: support -vos for coq >= 8.11 and default setting change

hendriktews:
This commit adds support for the new -vos compilation. For coq >=
8.11 only -vos can be used, depending on the config option
coq-compile-vos. For coq < 8.11 only -quick/-vio is used,
depending on option coq-compile-quick, as before. For a smooth
upgrade path, if coq-compile-vos has not been configured, the
users intention on whether to use -vos or not for coq >= 8.11 is
derived from coq-compile-quick.

Some defaults have been changed:

- parallel background compilation is the default now in case
  coq-compile-before-require is enabled.
- for coq < 8.11, quick/vio compilation with delayed vio-to-vo
  conversion is now the default

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -29,15 +29,22 @@ and the PG Trac http://proofgeneral.inf.ed.ac.uk/trac
 
 *** new menu Coq -> Auto Compilation for all background compilation options
 
+*** support for 8.11 vos compilation
+
+    See menu Coq -> Auto Compilation -> vos compilation, option
+    coq-compile-vos and subsection "11.3.3 Quick and inconsistent
+    compilation" in the Coq reference manual.
+
 *** support for 8.5 quick compilation
 
-    See new menu Coq -> Auto Compilation. Select "no quick" as
-    long as you have not switched to "Proof using" to compile
-    without -quick. Select "quick no vio2vo" to use -quick
-    without vio2vo (and guess what "quick and vio2vo" means ;-),
-    select "ensure vo" to ensure a sound development. See the
+    See new menu Coq -> Auto Compilation -> Quick compilation.
+    Select "no quick" as long as you have not switched to "Proof
+    using" to compile without -quick. Select "quick no vio2vo" to
+    use -quick without vio2vo (and guess what "quick and vio2vo"
+    means ;-), select "ensure vo" to ensure a sound development.
+    Quick compilation is only supported for Coq < 8.11. See the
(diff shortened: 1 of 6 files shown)


hendriktews:
@cpitclaudel @Matafou : I have not contributed for a while, I would therefore wait for your comments.
I decided to discontinue -quick/-vio and only support -vos for coq >= 8.11. I set the defaults such that parallel compilation is preferred and that -vos is not used, unless the user selects it or has configured a non-consistent compilation in coq-compile-quick before.
A documentation update will follow in a later commit.

Matafou:
Hi Hendrik, good to hear from you. Thanks a lot for your contributions.

In the long run we may want to use dune to manage all compilations, see #477. 

Are you also aware of coq_makefile produsing producing target  make foo.required_vos?

```
Typical compilation of a set of file using a build system.

Assume a file foo.v that depends on two files f1.v and f2.v. The command make foo.required_vos will compile f1.v and f2.v using the option -vos to skip the proofs, producing f1.vos and f2.vos. At this point, one is ready to work interactively on the file foo.v, even though it was never needed to compile the proofs involved in the files f1.v and f2.v.
```

hendriktews:
> In the long run we may want to use dune to manage all compilations, see #477.

From the answers in #477 I have the impression that this may take a while.

> Are you also aware of coq_makefile produsing producing target make foo.required_vos?

Sure. But this does not really help for auto compilation, does it?

Matafou:
> > Are you also aware of coq_makefile produsing producing target make foo.required_vos?
> 
> Sure. But this does not really help for auto compilation, does it?

true.

Shall we rebase/merge?

hendriktews:
Let me add a menu entry and update the documentation before merging.

Further, I tend to switch on automatic compilation by default. The original argument for keeping it off, namely that it conflicts with existing customization habits, is not true anymore for quite some time. Any opinion?

Matafou:
No opinion on this. I almost always have it on
.

cpitclaudel:
I've had it off for a while, since it used to not work with external OCaml modules.  I don't have a strong opinion, but I think it might surprise existing users to turn it on by default.

hendriktews:
Rebased and added menu, documentation and CHANGES. Automatic compilation is off by default as it used to be.

hendriktews:
> since it used to not work with external OCaml modules

Yes, this is because I have never looked into OCaml modules. I would be interested to work with somebody together, who wants to get this going.

PR #488, opened 2020-05-06
doc/ProofGeneral.texi: fix makeinfo

Flupp:
Fixes the following error:

    node `Coq Proof General' lacks menu item for `Proof using annotations' despite being its Up target


erikmd:
Good catch, thank you @Flupp!

This issue was also blocking the info manual build on MELPA…

So I added a CI test for the doc, better safe than sorry.

Issue #90, opened 2016-07-12
idtac output is getting parsed by PG as commands


jonleivent:
Goal True.
idtac "Require Dummy".

prints this, and messes up my PG windows:

-*- mode: compilation; -*-

echo "Require Dummy." > /tmp/ProofGeneral-coq17918FT.v /usr/local/bin/coqdep -Q /home/jil/coq/ "" /tmp/ProofGeneral-coq17918FT.v
*** Warning: in file /tmp/ProofGeneral-coq17918FT.v, library Dummy is required and has not been found in the loadpath!
*** Warning: in file /tmp/ProofGeneral-coq17918FT.v, library Dummy is required and has not been found in the loadpath!
/tmp/ProofGeneral-coq17918FT.vo /tmp/ProofGeneral-coq17918FT.glob /tmp/ProofGeneral-coq17918FT.v.beautified: /tmp/ProofGeneral-coq17918FT.v
/tmp/ProofGeneral-coq17918FT.vio: /tmp/ProofGeneral-coq17918FT.v

cpitclaudel:
Thanks a bug with compile-on-require, I think. @hendriktews ?

Matafou:
We had the same kind of problem with comments I remember.
P

Le mardi 12 juillet 2016, Clément Pit--Claudel <notifications@github.com> a
écrit :

hendriktews:
Yes, this looks like the regular expression that is searching for require commands is wrong.

hendriktews:
This is a duplicate of #75

hendriktews:
should be fixed by commit 3b621d71274bc62722876c7229c2e649be6cb378, please try.

PR #490, opened 2020-05-19
 Improve PG support of Show Proof Diffs

### LINKED ISSUE #487: Improve PG support of Show Proof Diffs
Currently, PG supports `Show Proof Diffs.` regarding Diffs' syntax highlighting, but the command itself has to be triggered manually.

ToDo:

* Create a new branch `coq-show-proof` from `master`
* Add a `defcustom coq-show-proof` option
* After each proof step (do nothing if outside of a proof):
    * If `coq-show-proof` is non-nil:
        * If `coq-diffs = nil`:
            * Send `Show Proof.` to the prover using `proof-shell-invisible-command`
        * If `coq-diffs = 'on`:
            * Send `Show Proof Diffs.` to the prover
        * If `coq-diffs = 'removed`:
            * Send `Show Proof Diffs removed.` to the prover
    * Then (normally, no code to write) the **\*response\*** buffer should be automatically selected.

* Add a menu checkbox for `coq-show-proof` − see e.g. #421:
    https://github.com/ProofGeneral/PG/blob/7501326a228a5d0af1ad1d5bef4e8640cbd1e43f/coq/coq-abbrev.el#L223

* When that menu checkbox is checked, if `coq-diffs = nil`, display a message in the minibuffer suggesting to enable `Coq > Diffs` as well.

* Bonus(**optional**): add a small test regarding the **\*response\*** text with `coq-show-proof = t` and `coq-diffs = nil`, on a proof such as `Goal True -> True. intros H.`

CyrilAnac:
Close #487


coq/coq-syntax.el
```diff
@@ -1297,6 +1297,9 @@ It is used:
   "*Font-lock table for Coq terms.")
 
 
+(defconst coq-show-proof-diffs-regexp
+  "^[ ]*Show Proof\\( Diffs\\| Diffs removed\\)?\\.$")
```
erikmd, line 1301:
```suggestion
  "\\`Show Proof\\(?: Diffs\\| Diffs removed\\)?\\.\\'")
```
Indeed
* `[ ]*` seems unnecessary as we just want to detect the "Show Proof…" commands that PG itself generates (so without leading spaces)
* the `\\(?:` marks the group "non-capturing"
* and finally \\` and \\' looks better to anchor the string

(according to https://www.gnu.org/software/emacs/manual/html_node/elisp/Syntax-of-Regexps.html#Syntax-of-Regexps )

**Note:** please re-test that regexp manually as well, with `(string-match-p coq-show-proof-diffs-regexp "…string…")`

generic/proof-config.el
```diff
@@ -399,6 +399,11 @@ NB: This setting is not used for matching output from the prover."
   :type 'regexp
   :group 'proof-script)
 
+(defcustom proof-show-proof-diffs-regexp nil
+  "Matches all Show Proof form"
```
erikmd, line 403:
```suggestion
  "Matches all "Show Proof" forms (specific to the Coq prover)."
```

generic/proof-shell.el
```diff
@@ -1126,7 +1126,7 @@ contains only invisible elements for Prooftree synchronization."
 	;; but we delay this until sending the next command, attempting
 	;; to parallelize prover and Emacs somewhat.  (PG 4.0 change)
 
-	(setq proof-action-list (cdr proof-action-list))
+  (setq proof-action-list (cdr proof-action-list))
```
erikmd, line 1129: You might want to revert that unneeded space change.

generic/proof-shell.el
```diff
@@ -1615,9 +1616,12 @@ by the filter is to send the next command from the queue."
       (setq proof-shell-delayed-output-end end)
       (setq proof-shell-delayed-output-flags flags)
       (if (proof-shell-exec-loop)
-	  (setq proof-shell-last-output-kind
+	  ;; (if (and  (string-match proof-show-diffs-regexp (car cmd))
+	  ;; 		  (memq 'empty-action-list flags))
+	  ;; (setq proof-shell-last-output-kind 'response)
```
erikmd, line 1621: Since that change is a no-op, you may want to add a commit that reverts the previous code (which was the same, without the comments)

generic/proof-shell.el
```diff
@@ -1712,7 +1716,7 @@ i.e., 'goals or 'response."
 	    (buffer-substring-no-properties rstart gmark)))
        ;; display goals output second so it persists in 2-pane mode
        (unless (memq 'no-goals-display flags)
-	 (pg-goals-display proof-shell-last-goals-output both))
+	 (pg-goals-display proof-shell-last-goals-output t))
```
erikmd, line 1719: I quickly tested locally and it seems it works with `both` instead of `t`. Can you please revert this and re-test?

coq/coq.el
```diff
@@ -1203,21 +1203,41 @@ be called and no command will be sent to Coq."
...
-    (list "Unset Silent." (if (coq--post-v810) (coq-diffs) "Show.")))
+    (list "Unset Silent." (if (coq--post-v810) (coq-diffs)
+                            (if coq-show-proof-stepwise
+    ((when (eq coq-diffs 'off)
+      "Show." "Show Proof." )
+    (when (eq coq-diffs 'on)
+      "Show." "Show Proof Diffs.")
+    (when (eq coq-diffs 'removed)
+       "Show." "Show Proof Diffs removed."))
+     "Show."))))
```
erikmd, line 1226: Some code is missing in the "then" case. At first sight (modulo indentation and parens' issues) you could write:

```suggestion
    (list "Unset Silent." (append (if (coq--post-v810) (coq-diffs) "Show.")
    (when coq-show-proof-stepwise
    (when (eq coq-diffs 'off) (list "Show Proof."))
    (when (eq coq-diffs 'on) (list "Show Proof Diffs."))
    (when (eq coq-diffs 'removed) (list "Show Proof Diffs removed.")))))
```

coq/coq.el
```diff
@@ -1203,21 +1203,41 @@ be called and no command will be sent to Coq."
...
     (if (coq--post-v810)
         (list "Unset Silent." (coq-diffs))
-      (list "Unset Silent.")))))
+      (list "Unset Silent.")))
+   ((or
+     ;; If doing (not closing) a proof, Show Proof if need be
+     (and (not (string-match-p coq-save-command-regexp-strict cmd))
+          (> (length coq-last-but-one-proofstack) 0)
+          (when coq-show-proof-stepwise
+            (list "Show." "Show Proof.")))))))
```
erikmd, line 1239: Given that `"Show."` prevents Diffs to show up, I suggest you only write:
```suggestion
            (list "Show Proof.")))))))
```

Also, the support of `Show Proof Diffs.` / `Show Proof Diffs removed` should be added here.

generic/proof-shell.el
```diff
@@ -1193,7 +1193,8 @@ contains only invisible elements for Prooftree synchronization."
 	     (or (null proof-action-list)
 		 (cl-every
 		  (lambda (item) (memq 'proof-tree-show-subgoal (nth 3 item)))
-		  proof-action-list)))))))
+		  proof-action-list)
+		 (string= (car (nth 1 item)) "Show.")))))))
```
erikmd, line 1197: Good point. At first sight you'll just have to replace this with something like:

```suggestion
		 (string-match-p proof-show-proof-diffs-regexp (last (nth 1 item)))))))))
```

(namely, we don't test that the "Show" (which is not inserted every time) is at the beginning (car), but that "Show Proof" is at the end (last)

coq/coq-syntax.el
CyrilAnac: \` doesn't seem to work, \b seems more appropriate

CyrilAnac:
@erikmd The feature seems to work now, I'm waiting for an other review or merging approve.

Matafou:
General question: what will happen if we switch to the "always silent. Do Show when needed" described here https://github.com/ProofGeneral/PG/pull/467#issuecomment-631345825?
Same question for the "Show goal Diff" feature.

erikmd:
Thanks @CyrilAnac ! the feature looks very nice.

However it seems the following bug is introduced:

(1) when opening a .v file

```coq
Lemma test : forall n : nat, n = n.
  reflexivity. (* (2) then doing "C-c C-RET" to here *)
```

(3) and finally doing "C-c C-u" to backtrack, we get:

```
condition-case: Wrong type argument: stringp, nil
```

− I'm using Coq 8.11.1 on GNU/Linux

erikmd:
Hi @Matafou 
> General question: what will happen if we switch to the "always silent. Do Show when needed" described here [#467 (comment)](https://github.com/ProofGeneral/PG/pull/467#issuecomment-631345825)?
> Same question for the "Show goal Diff" feature.

From my understanding the features are not incompatible but essentially orthogonal.

To summarize the situation:

* This PR #490 just amounts to append (if the corresponding option is enabled) a `Show Proof…` command at the end of the queue, typically after `Show.` if there is one, or just in the queue alone otherwise.
So if the `Show` command must be triggered systematically to display goals (as per #188), we will just have the simplification that PG always append `(list "Show." "Show Proof…")` (assuming option `coq-show-proof-stepwise` is enabled)

* The PR #421 that added `Coq Diffs` to goals relies on the Coq option `Set Diffs "on".` that is not synchronized, so to avoid the fact that the option would be reset at each backtrack, the implemented solution is to re-send a `Set Diffs "on".` after each backtracking. It happens that **if we do `Show.` at that point, the diff colors are not shown** because the following feature has not yet been integrated in Coq: https://github.com/coq/coq/pull/10020#issuecomment-557092744 − Fortunately, the `Set Diffs "on".` itself triggers a re-display of the goal if there is some, so actually this plays the role of the skipped `Show.` :)

coq/coq-syntax.el
```diff
@@ -1297,6 +1297,9 @@ It is used:
   "*Font-lock table for Coq terms.")
 
 
+(defconst coq-show-proof-diffs-regexp
+  "\\bShow Proof\\(?: Diffs\\| Diffs removed\\)?\\.\\b")
```
erikmd, line 1301: the [`\\b`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Regexp-Backslash.html#Regexp-Backslash) token seems a bit off-topic here:

```suggestion
  "\\`Show Proof\\(?: Diffs\\| Diffs removed\\)?\\.\\'")
```

coq/coq.el
```diff
@@ -160,6 +160,16 @@ See also option `coq-hide-additional-subgoals'."
   :type '(choice regexp (const nil))
   :group 'coq)
 
+(defcustom coq-show-proof-stepwise nil
+  "Display the proof terms stepwise in the *response* buffer.
+  This option can be combined with option `coq-diffs'.
+  It is mostly useful in three window mode, see also
+  `proof-three-window-mode-policy' for details."
```
erikmd, line 167: You should probably remove the leading spaces here:

```suggestion
This option can be combined with option `coq-diffs'.
It is mostly useful in three window mode, see also
`proof-three-window-mode-policy' for details."
```

(otherwise the output of <kbd>C-h v coq-show-proof-stepwise RET</kbd> is weird)

erikmd:
> Thanks @CyrilAnac ! the feature looks very nice.
> 
> However it seems the following bug is introduced:
> 
> (1) when opening a .v file
> 
> ```coq
> Lemma test : forall n : nat, n = n.
>   reflexivity. (* (2) then doing "C-c C-RET" to here *)
> ```
> 
> (3) and finally doing "C-c C-u" to backtrack, we get:
> 
> ```
> condition-case: Wrong type argument: stringp, nil
> ```
> 
> − I'm using Coq 8.11.1 on GNU/Linux

Apparently this can be reproduced by directly processing the first line of `test_stepwise.v` till the end of a comment. I added a regression test (that will fail for now)

coq/coq.el
```diff
@@ -1203,21 +1213,45 @@ be called and no command will be sent to Coq."
...
      ;; silent (there is no goal to show). Still, we need to "Set Diffs" again
      (string-match-p "BackTo\\s-" cmd))
     (if (coq--post-v810)
-        (list "Unset Silent." (coq-diffs))
-      (list "Unset Silent.")))))
+        (list "Unset Silent." (coq-diffs) )
+      (list "Unset Silent.")))
+   ((or
+     ;; If starting a proof, Show Proof if need be
+     (coq-goal-command-str-p cmd)
```
erikmd, line 1245: FYI I adeed this line, as otherwise after doing `(setq coq-show-proof-stepwise t)` and starting a proof, the proof term was not shown after processing the first line of the theorem (`Lemma … : …`)

I've just seen that this function has the following comment:

https://github.com/ProofGeneral/PG/blob/71748af5bc0f813d4c6a42f303d2c2cec85b3c9a/coq/coq-syntax.el#L1027-L1030

so are you OK all the same @Matafou with using it just once? basically even if the `coq-goal-command-str-p` does not perfectly detects all goal starts (you may speak of a heuristic), IMHO it is needed here to have a fully complete feature #487 …

Also, @CyrilAnac in you next commit, can you reindent the whole function `coq-empty-action-list-command` ? (basically by selecting all the code of the function and typing <kbd>TAB</kbd>)

coq/coq.el
```diff
@@ -1203,21 +1213,45 @@ be called and no command will be sent to Coq."
      ;; If user issued a printing option then t printing.
      (and (string-match-p "\\(S\\|Uns\\)et\\s-+Printing" cmd)
           (> (length coq-last-but-one-proofstack) 0)))
-    (list "Show."))
+    (list "Show."
+    (when coq-show-proof-stepwise
+        (or
+    (when (eq coq-diffs 'off) "Show Proof.")
```
erikmd, line 1219: @CyrilAnac last remark: could you add at the end of the *docstring* of this function, a remark such as the following?

```
Note: the last command added if `coq-show-proof-stepwise' is set
should match the `coq-show-proof-diffs-regexp'."
```

as it looks important to document this property somewhere (e.g. in the docstring)

coq/coq-syntax.el
CyrilAnac: This suggested regex doesn't work for me, it doesn't match `Show Proof.` for example. I don't know why.

erikmd: @CyrilAnac can you give more details? what do you mean by "doesn't work"? On my setup (emacs 26.1, GNU/Linux), this regex indeed returns a non-nil value:

```
(string-match-p "\\`Show Proof\\(?: Diffs\\| Diffs removed\\)?\\.\\'" "Show Proof.")
```

CyrilAnac: Oh, right. I didn't understand the point of these anchors. With this test the result is 0 like expected.

CyrilAnac:
> > Thanks @CyrilAnac ! the feature looks very nice.
> > However it seems the following bug is introduced:
> > (1) when opening a .v file
> > ```coq
> > Lemma test : forall n : nat, n = n.
> >   reflexivity. (* (2) then doing "C-c C-RET" to here *)
> > ```
> > 
> > 
> > (3) and finally doing "C-c C-u" to backtrack, we get:
> > ```
> > condition-case: Wrong type argument: stringp, nil
> > ```
> > 
> > 
> > − I'm using Coq 8.11.1 on GNU/Linux
> 
> Apparently this can be reproduced by directly processing the first line of `test_stepwise.v` till the end of a comment. I added a regression test (that will fail for now)

@erikmd Indeed, I introduced a bug [here](https://github.com/ProofGeneral/PG/pull/490/commits/133a11f0df7dd83848a34ff680b38d73d6b57804#diff-85331fb300ef6b347a5f53398fce91b4R1198), the result of `(car (nth 1 (car (last proof-action-list))))` is `nil` and the `string-match-p` throw a wrong type argument error.
The incoming commit should fix this issue.

coq/coq.el
CyrilAnac: Yes I will 👍

erikmd:
Also before merging this feature, it'd be great to have users test it a little bit further (notably to ensure there is no unforeseen regression), maybe @jfehrle or @anton-trunov would have the time to do this?

FYI to retrieve the code of this PR it should suffice to uninstall `proof-general` from MELPA and follow https://github.com/ProofGeneral/PG#using-git-manual-compilation-procedure then do:

```
cd ~/.emacs.d/lisp/PG
git config --global alias.pr '!f() { if [ $# -lt 1 ]; then echo "Usage: git pr <id> [<remote>]  # assuming <remote>[=origin] is on GitHub"; else git checkout -q "$(git rev-parse --verify HEAD)" && git fetch -fv "${2:-origin}" pull/"$1"/head:pr/"$1" && git checkout pr/"$1"; fi; }; f'
git pr 490
make clean
# make  # optional
```

and the two options at stake are: `Coq > Diffs > on` and `Coq > Show Proof (Diffs)`, see screenshot below.

2020-05-25_12-31-01_Screenshot_Show_Proof

erikmd:
Sorry, the feature is not yet ready.

> BTW if you have some time, you might want to add one or two unit tests (?)

we definitely should add such tests!

I've spotted the following two bugs:

1. If we enable "Coq Diffs" and "Show Proof", with Coq 8.10, interaction is broken (because Coq Diffs are available in Coq 8.10, but *not Show Proof Diffs*, which have been introduced in https://github.com/coq/coq/pull/10494/commits/5ec7eca640a6ee495eb2c0fb8d8a4076256ff96d) → Cyril could you add some appropriate conditions `(if (coq--post-v811) …)`? (I've updated the CI configuration to also test with Coq 8.9 and Coq 8.10, but we should also add some tests that make the `ci/test.sh` script fail with Coq 8.10)

2. If "Coq Diffs" and "Show Proof" are disabled and we backtrack (e.g. with Coq 8.10) then run `C-c C-n`, then the frontend loops…

jfehrle:
>  Coq Diffs are available in Coq 8.10, but not Show Proof Diffs, which have been introduced in coq/coq@5ec7eca) → Cyril could you add some appropriate conditions (if (coq--post-v811) …)

Show Proof Diffs is in 8.11.  Perhaps you meant "if (coq--post-v810)"?

> maybe @jfehrle or @anton-trunov would have the time to do this?

I don't use PG or emacs on a regular basis.  Looks like I don't even have emacs installed on my new laptop.  Perhaps Anton would be willing to try it first.

erikmd:
Hi @jfehrle, thanks for your comment.

> Perhaps you meant "if (coq--post-v810)"?

nope: the `(coq--post-v811)` is actually a `≥` :)

https://github.com/ProofGeneral/PG/blob/8273e22e6da7b20e51fa330116bee20964d5dab8/coq/coq-system.el#L200-L202

ci/coq-tests.el
```diff
@@ -110,8 +110,13 @@
    #'proof-done-invisible
    'no-error-display 'no-response-display 'no-goals-display))
 
+(defun coq-set-flags (val flags)
+  (when (member 'show-proof-stepwise flags) (setq coq-show-proof-stepwise val))
+  (when (member 'diffs-on flags) (if val (setq coq-diffs 'on) (setq coq-diffs 'off)))
+  )
```
erikmd, line 116: Nitpick: the closing `)` should be merged with the last line.

ci/coq-tests.el
```diff
@@ -110,8 +110,13 @@
...
 
-(defun coq-fixture-on-file (file body)
+
+(defun coq-fixture-on-file (file  body &rest flags)
```
erikmd, line 119: nitpick: you have some unneeded double-space here, `file  body`

ci/coq-tests.el
```diff
@@ -166,12 +173,18 @@ For example, COMMENT could be (*test-definition*)"
                   (with-current-buffer "*response*"
                     (buffer-substring-no-properties (point-min) (point-max)))))))
 
+(defun coq-should-buffer (message)
```
erikmd, line 176: you may want to name that variable `regexp` for example, instead of `message`?

ci/coq-tests.el
```diff
@@ -180,19 +193,30 @@ For example, COMMENT could be (*test-definition*)"
...
+   (coq-test-full-path "test_stepwise.v") 
    (lambda ()
      (coq-test-goto-before "(*test-definition*)")
      (proof-goto-point)
      (proof-shell-wait)
      (coq-should-response "trois is defined"))))
 
 
+(ert-deftest 021_coq-test-regression-goto-point ()
+  "Regression test for proof-goto-point after a comment, PR #90"
```
erikmd, line 206: I think I had written that docstring, but it should probably be `#490` instead of `#90`

ci/coq-tests.el
```diff
@@ -180,19 +193,30 @@ For example, COMMENT could be (*test-definition*)"
...
+       (proof-goto-point)
+       (proof-shell-wait)
+       t)))
+
+
 (ert-deftest 030_coq-test-position ()
   "Test locked region after Qed."
-  (coq-fixture-on-file
-   (coq-test-full-path "test_stepwise.v")
+  (coq-fixture-on-file
```
erikmd, line 218: please remove the trailing spaces you added after each `(coq-fixture-on-file ` and `"test_stepwise.v") `, see URL https://github.com/ProofGeneral/PG/pull/490/files in order to spot them.

ci/coq-tests.el
erikmd: and you could rename that function to `coq-should-buffer-regexp`, then define:

```
(defun coq-should-buffer-string (str)
  "Particular case of `coq-should-buffer-regexp'."
  (coq-should-buffer-regexp (regexp-quote str)))
```

ci/coq-tests.el
```diff
@@ -245,15 +269,43 @@ For example, COMMENT could be (*test-definition*)"
...
+
+(ert-deftest 080_coq-test-regression-show-proof-stepwise()
+  "Regression test for the \"Show Proof\" option"
+    (coq-fixture-on-file 
+   (coq-test-full-path "test_stepwise.v") 
+   (lambda ()
+     (coq-test-goto-before " (*test-insert*)")
+     (proof-goto-point)
+     (proof-shell-wait)
+     (coq-should-response "(fun (A : Prop) (proof_of_A : A) => ?Goal)"))
```
erikmd, line 290: then maybe you could use here `(coq-should-buffer-string "(fun …` (see previous review comment)

ci/coq-tests.el
```diff
@@ -245,15 +269,43 @@ For example, COMMENT could be (*test-definition*)"
...
+  "Test for Show Proof Diffs"
+    (coq-fixture-on-file 
+   (coq-test-full-path "test_stepwise.v") 
+   (lambda ()
+     (coq-test-goto-before " (*test-insert*)")
+     (proof-goto-point)
+     (proof-shell-wait)
+     ;; If coq--post-v811, it should be "Show Proof Diffs." otherwise "Show Proof."
+     (if (coq--post-v811)
+	 (coq-should-buffer  "<diff\\.added\\.bg>(fun <diff\\.added>(</diff\\.added>A : Prop<diff\\.added>) (proof_of_A : A)</diff\\.added> => \\?Goal)</diff\\.added\\.bg>")
```
erikmd, line 304: Likewise, this could be something like:

`(coq-should-buffer-string "<diff.added.bg>(fun <diff.added>(</diff.added>A : Prop<diff.added>) (proof_of_A : A)</diff.added> => ?Goal)</diff.added.bg>")`

coq/coq.el
```diff
@@ -1203,21 +1215,63 @@ be called and no command will be sent to Coq."
...
      (and (string-match-p "BackTo\\s-" cmd)
           (> (length coq-last-but-one-proofstack) coq--retract-naborts)))
     ;; "Set Diffs" always re-prints the proof context with (if enabled) diffs
-    (list "Unset Silent." (if (coq--post-v810) (coq-diffs) "Show.")))
+    ;; (list "Unset Silent." (if (coq--post-v810) (coq-diffs) "Show.")
+    ;;       (when coq-show-proof-stepwise
+    ;;         (or
+    ;;          (when (eq coq-diffs 'off) "Show Proof.")
+    ;;          (when (eq coq-diffs 'on) "Show Proof Diffs.")
+    ;;          (when (eq coq-diffs 'removed) "Show Proof Diffs removed.")))))
```
erikmd, line 1241: now you can probably remove these commented lines.

Matafou:
LGTM

PR #495, opened 2020-05-27
New hook for early prompt/output analyzis.

Matafou:
proof-state-change-pre-hook happens earlier than
proof-state-change-hook, i.e. before proof-done-advancing. This should
be used to register information in the currently processed span before
proof-done-advancing classifies it.

Historically PG design was to gather these information during
proof-done-advancing (or in its hook called at the end) by just
looking at the command statement. But it is often useful to look at
the output (messages and/or prompt) to gather more accurate
information. Some of this information may be needed DURING
proof-done-advancing. Hence this early hook.

### DIFF
--- a/coq/coq.el
+++ b/coq/coq.el
@@ -661,7 +661,7 @@ If locked span already has a state number, then do nothing. Also updates
 ;; This hook seems the one we want.
 ;; WARNING! It is applied once after each command PLUS once before a group of
 ;; commands is started
-(add-hook 'proof-state-change-hook #'coq-set-state-infos)
+(add-hook 'proof-state-change-pre-hook #'coq-set-state-infos)
 
 
 (defun count-not-intersection (l notin)
--- a/doc/PG-adapting.texi
+++ b/doc/PG-adapting.texi
@@ -3107,7 +3107,8 @@ before and after sending the command.
 In case @var{cmd} is (or yields) nil, do nothing.
 
 @var{invisiblecallback} will be invoked after the command has finished,
-if it is set.  It should probably run the hook variables
+if it is set. It should probably run the hook variables
+@samp{@code{proof-state-change-pre-hook}} and
 @samp{@code{proof-state-change-hook}}.
 
 @var{flags} are additional flags to put onto the @samp{@code{proof-action-list}}.
@@ -4056,8 +4057,9 @@ not need to use these directly.
 @c TEXI DOCSTRING MAGIC: proof-grab-lock
 @defun proof-grab-lock &optional queuemode
 Grab the proof shell lock, starting the proof assistant if need be.@*
-Runs @samp{@code{proof-state-change-hook}} to notify state change.
-If @var{queuemode} is supplied, set the lock to that value.
+Runs @samp{@code{proof-state-change-pre-hook}} and
+@samp{@code{proof-state-change-hook}} to notify state change. If
+@var{queuemode} is supplied, set the lock to that value.
 @end defun
 
 @c TEXI DOCSTRING MAGIC: proof-release-lock
--- a/easycrypt/easycrypt-hooks.el
+++ b/easycrypt/easycrypt-hooks.el
@@ -68,7 +68,7 @@
        (setq easycrypt-proof-weak-mode (car (cdr infos)))
      )))
 
-(add-hook 'proof-state-change-hook 'easycrypt-set-state-infos)
+(add-hook 'proof-state-change-pre-hook 'easycrypt-set-state-infos)
 
 (defun easycrypt-find-and-forget (span)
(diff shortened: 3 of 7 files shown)


Matafou:
Trying to be a good guy proceeding with PR instead of wild (and buggy commit). Waiting for tests and comments.

When trying to correct PR #467 I detected that there was a bit of dependency loop between proof-done-advancing and its end-hook. Some of  the information gathered in the span by this hook (e.g. is this command opening a goal?) was actually needed by proof-done-advancing (e.g.should we increment proof-nesting-depth?).

Notably this does actually not help to correct PR #467 since the information (are we inside a proof?) is needed even earlier. This is the subject of another commit.

erikmd:
Hi @Matafou, thanks, the idea looks good; two questions:

* which kind of "typical tests" would you have in mind? (BTW it might be possible to automate part of these tests in the CI if you'd like)

* shouldn't these lines be edited as well?
    * https://github.com/ProofGeneral/PG/blob/2ada467c2c2964a605659714f2a267db2c915952/generic/proof-script.el#L1438
    * https://github.com/ProofGeneral/PG/blob/632a3d7f9ded16faaf58e1c0769bcd4f7c8193e3/generic/proof-tree.el#L817
    * https://github.com/ProofGeneral/PG/blob/733cd24a7368ee186884da488da0f59bbedb627e/easycrypt/easycrypt-hooks.el#L71

Matafou:
> Hi @Matafou, thanks, the idea looks good; two questions:
> 
> * which kind of "typical tests" would you have in mind? (BTW it might be possible to automate part of these tests in the CI if you'd like)

Difficult to test this currently without writing some lisp testing. Maybe once the proof-nesting generic code works (although it is more accurate now, currently the `proof-nesting-depth` variable is incorrect when backtracking).

> * shouldn't these lines be edited as well?
>   
>   * https://github.com/ProofGeneral/PG/blob/2ada467c2c2964a605659714f2a267db2c915952/generic/proof-script.el#L1438

No. The proof-done-advancing-hook remains. It is only that some of its content were triggered to late, hence the need for another hook.

>   * https://github.com/ProofGeneral/PG/blob/632a3d7f9ded16faaf58e1c0769bcd4f7c8193e3/generic/proof-tree.el#L817

Well spotted. I will add it there as the information about state number is typically done by the pre-hook now. It would probably break proof-tree otherwise.

>   * https://github.com/ProofGeneral/PG/blob/733cd24a7368ee186884da488da0f59bbedb627e/easycrypt/easycrypt-hooks.el#L71

This will have no effect but it is more consistent to change this to pre-hook too.

Thanks!

Matafou:
I just had a look at the test infrastructure. Can we easily add adhoc lisp code to a test?

erikmd:
> I just had a look at the test infrastructure. Can we easily add adhoc lisp code to a test?

Yes, it suffices to add an ERT form in `coq-tests.el` such as :

```elisp
(ert-deftest 090_coq-test-pr-495()
  "Test for PR #495 about …" 
  (coq-fixture-on-file (coq-test-full-path "test_pr495.v") 
  (lambda ()
  ;elisp code, e.g.:
  (proof-process-buffer)
  (proof-shell-wait)
  ;optional assertion: (should…)
  ;otherwise: t)
```

assuming you need an accompanying file `test_pr495.v`

Then you can run the tests in two alternative ways:

1. `./ci/test.sh` or:
2. <kbd>C-x C-f coq-tests.el RET</kbd>
    <kbd>M-: (progn (load-file "coq-tests.el") (call-interactively #'ert)) RET</kbd>

Matafou:
Cool. I will try to adopt this systematically.

erikmd:
If you don't need an ad-hoc file, you can just replace `(coq-test-full-path "test_pr495.v")` with `nil` and the `coq-fixture-on-file` function will take care of starting a fresh Coq instance on a *temporary* Coq file

erikmd:
> Then you can run the tests in two alternative ways:
> 
>     1. `./ci/test.sh` or:
> 
>     2. C-x C-f coq-tests.el RET
>        M-: (progn (load-file "coq-tests.el") (call-interactively #'ert)) RET

or more simply, just evaluating Line 8 in a buffer:
https://github.com/ProofGeneral/PG/blob/d20ad3dde9be1fc9702f794e89a127f3c3b312fd/ci/coq-tests.el#L6-L8

erikmd:
Finally in the case of a failure in ERT interactive mode, you can click on a failing test and press `b` to get a backtrace, see also:
https://www.gnu.org/software/emacs/manual/html_node/ert/Running-Tests-Interactively.html#Running-Tests-Interactively

Matafou:
I merge this and will add tests later.

Issue #493, opened 2020-05-26
Test files should not provide features


tarsius:
Test files should not `provide` a feature. Providing a feature is only suitable for libraries that are loaded by other libraries/packages or by the end user using `require`. For `require` to be able to do its job, the library that provides the feature has to be located on the `load-path`, but test files should *not* be located on the `load-path`.

You are probably already loading this test file using something `(load "relative/path/to/it")` and can just remove the `provide` form.  If not, then you probably just have to replace a `require` with a such a `load` form somewhere.

One reason why test files should not `provide` a feature and/or reside on the `load-path` is that authors tend to name those files just `tests.el` or something like that.  If only one package does that, then that works, but it only two out of several thousands and things start to get problematic. In fact I usually don't notice such issues until there actually are two packages that provide the exact same "testing feature".

In this case the other package is `learn-ocaml`. @erikmd, you added `init-tests.el` in both cases, could you please address this request. Thanks!

<sub>I have written similar messages many many times so I now [use a template](https://help.github.com/en/github/writing-on-github/working-with-saved-replies). I try to adjust it to match each individual case but it might still not apply 100% in all cases.</sub>

tarsius:
Thanks!

erikmd:
Thanks a lot @tarsius for reporting this and for your detailed explanation!

Fixed in PG master, will fix learn-ocaml master as well.

hendriktews:
This probably also applies to coq/coq-par-test.el

PR #496, opened 2020-05-28
fix: test files should not provide features

erikmd:
this PR is a follow-up of https://github.com/ProofGeneral/PG/issues/493#issuecomment-634493988


erikmd:
The CI is green so I'll merge this tonight if there is no objection.

Issue #514, opened 2020-10-06
Proof process hang with bad tactics in Coq


mgree:
When running the following script:

```Coq
Theorem thm : True.
Proof. assert True.{ reflexivity. } apply H. Qed.
```

The proof process will hang when it gets to the first opening brace. I'm able to break things with `C-c C-x`, but some students have reported a general Emacs hang.

I'm running Emacs 26.3, Coq 8.10.2 (students are running 8.12.0), and ProofGeneral 20200623.1748. I'm on macOS 10.13.6.

cpitclaudel:
The general Emacs hang seems surprising; do you / they have a repro?  Otherwise, yes, there are many known ways to confuse the REPL, including defining notations that include periods or sending things to Coq that looks like sentence endings to Proof General.  Migrating away from the REPL is the only robust solution here, I think.

mgree:
I'm trying to work with students to get a repro, but (a) it's challenging with novices (this is a CS 2 course!) and (b) it's even more challenging with remote instruction. It could be that they're not aware of appropriate ways to break from the prover loop.

My understanding is that there are some longer term plans to switch to a better protocol... do you know where I can track that progress?

cpitclaudel:
> It could be that they're not aware of appropriate ways to break from the prover loop.

I *think* pressing C-c C-c should break out of that loop (:crossed_fingers:)

> My understanding is that there are some longer term plans to switch to a better protocol... do you know where I can track that progress?

We don't really have progress being tracked, unfortunately

hendriktews:
Hi,

your line

> Proof. assert True.{ reflexivity. } apply H. Qed.

is syntactically not correct, both coqtop and coqc report an
error on it (just recording this here for those for which this is
not obvious).

Proof General hangs, because the Proof General parser decides to
send `assert True.{` as one command to coqtop. For coqtop
however, this is not a complete command, so it doesn't send a
reply and then Proof General is waiting for ever. Emacs stays
responsive, but the communication with Coq is in a deadlock.

Instead of C-c C-x, which kills the prover, you can also try C-c
C-c, which only sends an interrupt. This is also in the tool-bar
and in the Proof General menu (Interrupt).

I agree that it would be nice to get a decent error in Proof
General for this example. However, I am not sure what could be
done. We might be able to fix the regular expressions such that
for this example Proof General sends a command that directly
gives an error. But there will always remain examples that the
regular expressions won't parse correctly... Therefore I don't
[...]

PR #516, opened 2020-10-12
Optional timeout warning when waiting for the proof shell

ibynx:
With bad tactics in Coq (#514), Proof General will hang forever (but can be interrupted with `C-c C-c` or `C-c C-x`). In response to @hendriktews's workaround suggestion, this PR exposes a variable `proof-shell-timeout-warn` that, if an integer, enables a minibuffer warning after waiting on the proof shell for more than `proof-shell-timeout-warn` seconds. 

It doesn't fix the core issue, but the minibuffer warning will remind users of using `C-c C-c` and `C-c C-x` to cancel scripting. 

The text of the warning is "This command is taking a while. Is it malformed? Do C-c C-c or C-c C-x to abort." Default timeout is 30 seconds.

### DIFF
--- a/generic/proof-script.el
+++ b/generic/proof-script.el
@@ -2029,9 +2029,9 @@ This function expects the buffer to be activated for advancing."
 
   ;; arm the timeout timer
   ;; cancelled in proof-shell-exec-loop unless proof-shell-busy
-  (if proof-shell-timeout-warn-p
+  (if proof-shell-timeout-warn
       (setq proof-shell-timer
-          (run-with-timer proof-shell-timeout-warn-length nil
+          (run-with-timer proof-shell-timeout-warn nil
                           'message "This command is taking a while. \
 Is it malformed? Do C-c C-c or C-c C-x to abort."))))
 
--- a/generic/proof-script.el
+++ b/generic/proof-script.el
@@ -2025,7 +2025,15 @@ This function expects the buffer to be activated for advancing."
 	(lastpos   (nth 2 (car semis)))
 	(vanillas  (proof-semis-to-vanillas semis displayflags)))
     (proof-script-delete-secondary-spans startpos lastpos)
-    (proof-extend-queue lastpos vanillas)))
+    (proof-extend-queue lastpos vanillas))
+
+  ;; arm the timeout timer
+  ;; cancelled in proof-shell-exec-loop unless proof-shell-busy
+  (if proof-shell-timeout-warn-p
+      (setq proof-shell-timer
+          (run-with-timer proof-shell-timeout-warn-length nil
+                          'message "This command is taking a while. \
+Is it malformed? Do C-c C-c or C-c C-x to abort."))))
 
 (defun proof-retract-before-change (beg end)
   "For `before-change-functions'.  Retract to BEG unless BEG and END in comment.
--- a/generic/proof-shell.el
+++ b/generic/proof-shell.el
@@ -149,6 +149,12 @@ This flag is set for the duration of `proof-shell-kill-function'
 to tell hooks in `proof-deactivate-scripting-hook' to refrain
 from calling `proof-shell-exit'.")
 
+(defvar proof-shell-timer nil
+  "A timer that alerts the user when the current command sent to the
+shell is taking too long and might be malformed. This is reset on entry
+of `proof-shell-exec-loop' and set when the next command is sent to
+the process. Disable by setting `proof-shell-timeout-p' to nil.
+Configure with `proof-shell-timeout-length'")
 
 
 ;;
@@ -1176,8 +1182,8 @@ contains only invisible elements for Prooftree synchronization."
 	  (proof-shell-handle-error-or-interrupt 'interrupt flags))
 
 	(if proof-action-list
-	    ;; send the next command to the process.
-	    (proof-shell-insert-action-item (car proof-action-list)))
+	  ;; send the next command to the process.
+	  (proof-shell-insert-action-item (car proof-action-list)))
 
 	;; process the delayed callbacks now
 	(mapc 'proof-shell-invoke-callback cbitems)
--- a/generic/proof-shell.el
+++ b/generic/proof-shell.el
@@ -1181,9 +1181,9 @@ contains only invisible elements for Prooftree synchronization."
 	  (setq cbitems nil)
 	  (proof-shell-handle-error-or-interrupt 'interrupt flags))
 
-	(if proof-action-list
-	  ;; send the next command to the process.
-	  (proof-shell-insert-action-item (car proof-action-list)))
+  (if proof-action-list
+	    ;; send the next command to the process.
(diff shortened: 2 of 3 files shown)


Matafou:
Sorry for the delay, I will review your PR shortly. Thanks for it!

generic/proof-shell.el
```diff
@@ -1175,7 +1187,7 @@ contains only invisible elements for Prooftree synchronization."
 	  (setq cbitems nil)
 	  (proof-shell-handle-error-or-interrupt 'interrupt flags))
 
-	(if proof-action-list
+  (if proof-action-list
```
Matafou, line 1190: Incorrect indentation.

generic/proof-script.el
```diff
@@ -2025,7 +2025,16 @@ This function expects the buffer to be activated for advancing."
...
+    (proof-extend-queue lastpos vanillas))
+
+  ;; arm the timeout timer
+  ;; cancelled 1. in `proof-shell-exec-loop' unless `proof-shell-busy' or
+  ;; 2. in the case of error, in `proof-shell-error-or-interrupt-action'
+  (if proof-shell-timeout-warn
+      (setq proof-shell-timer
+          (run-with-timer proof-shell-timeout-warn nil
+                          'message "This command is taking a while. \
+Is it malformed? Do C-c C-c to interrupt prover or C-c C-x to terminate it."))))
```
Matafou, line 2037: For the message, please use something like
```
(substitute-command-keys "bla bla \\[proof-interrupt-process] bla bla")
```
This way the actual shortcut will be displayed, even if it has changed from the default one.

generic/proof-shell.el
```diff
@@ -738,6 +745,11 @@ unless the FLAGS for the command are non-nil (see `proof-action-list')."
     ;; proof-action-list is empty on error.
     (setq proof-action-list nil)
     (proof-release-lock)
+    (unless proof-shell-busy
```
Matafou, line 748: I am not sure this test is necessary. Why would the shell be busy if an error is detected? Probably not harmful.

Matafou (commented):
General remarks. Other remarks in the code.

- You should probably not make a PR from your own "master" branch. You should define a dedicated branch in you repository instead. No big deal.
- While testing I noticed that the message is displayed sometimes apparently for no reason. It seems this happens when PG processes comments. Maybe because in this cases PG do not send anything to coq actually, and this gets no response either. Probably a missing cancel-timer somewhere. I will investigate a bit more.
This almost good I would say.

Matafou:
> * Probably a missing cancel-timer somewhere. I will investigate a bit more.

It is just that the timer should not be launched when the semi's type is `'comment`.

generic/proof-shell.el
andyqhan: I left it in, but can remove if you'd like.

andyqhan:
@Matafou I've implemented your suggestions — thank you so much for the review! 

> * You should probably not make a PR from your own "master" branch. You should define a dedicated branch in you repository instead. No big deal.

My bad, I'll make sure to do that in the future. 

> * While testing I noticed that the message is displayed sometimes apparently for no reason. It seems this happens when PG processes comments. Maybe because in this cases PG do not send anything to coq actually, and this gets no response either. Probably a missing cancel-timer somewhere. I will investigate a bit more.

I changed the conditional for arming the timer to:
```
(if (and proof-shell-timeout-warn
         (not (eq (caar semis) 'comment)))
```

I'm not sure whether this is what you had in mind.

hendriktews:
Thanks for your contribution!
In #514 I described an approach that sets and clears the timer for each command sent to Coq. Instead, your approach is to set and clear the timer only once for each asserted region. This is also possible. However, you need to consider two points.
1. Comments, as pointed out already by @Matafou: If the asserted region consists only of comments, then `proof-add-to-queue` might process them without calling `proof-shell-exec-loop`. Checking whether the first element is a comment (as you do now) is not good enough, because the asserted region may consist of a mixture of comments and real commands. If you want to stick to your approach, I suggest to move setting up the timer to `proof-add-to-queue` after the leading comments have been processed there.
2. quickly asserting twice: I believe it is possible to assert another region while the previous region is still processing. In this case, you will never kill the timer for the first region.

generic/proof-shell.el
```diff
@@ -1190,6 +1202,12 @@ contains only invisible elements for Prooftree synchronization."
 	    (pg-processing-complete-hint))
 	  (pg-finish-tracing-display))
 
+  (unless proof-shell-busy
+		;; if the shell isn't still busy, cancel timer
+		(if (and proof-shell-timer proof-shell-timeout-warn)
+			(progn (cancel-timer proof-shell-timer)
+			       (setq proof-shell-timer nil))))
+
```
hendriktews, line 1210: The indentation seems wrong here. Would it be possible to fix that?

andyqhan:
Thank you for the speedy review. 

I'm reconsidering the need for this PR, because I can't seem to get the MWE given in #514 to hang (probably because of the fixes in 0fdb1ae633baeb9afb07bbd8632bece5976f95f2, I'm getting `Syntax Error: Lexer: Undefined token`). However, if the maintainers still think there's a need, I'd be happy to continue working on this feature — it's just that not being able to reproduce the hanging anymore means that I have trouble working on it further.

Matafou:
If you want a script where coq still hangs you can use this:
```
Lemma foobar: True.
Proof.
  idtac.+ }
```

andyqhan:
Ok, I believe I've fixed the issues. I'm not quite sure what to do with the case when someone can assert quickly multiple times in a row (the timer won't start if it's already running — that may help), but it now doesn't start the timer if the processed region was all comments. In order to do this, per @hendriktews 's suggestion, I've moved setting the timer to `proof-add-to-queue`, and arm the timer only when the length of `proof-shell-slurp-comments` is less than the length of `proof-action-list` (that is, when the region isn't only comments). Could someone take a look?

Matafou:
It seems to work, but is it really what we expect?
When one launch a big thunk of coq commands, if the global execution of the whole thunk takes more that 30sec we get the message. One would expect that we put a timer on each command. No?

hendriktews:
> It seems to work, but is it really what we expect?

No. It seems that my comment from March 21, that the approach followed here is also possible, was wrong. I would suggest to follow my outline in #514 instead.

andyqhan:
I moved the timer setup to `proof-shell-insert`, which should mean that it is armed per command and not per region. I tried putting it in `proof-shell-exec-loop` like hendriktews suggested, but it seems that if a command hangs, that function isn't called. In order to deal with the fact that a region can take more than `proof-shell-timeout-warn` seconds to complete, I cancel the timer before setting it if the timer exists. Also, I'm checking that the argument of `proof-shell-insert` is a visible command by checking that the callback is `'proof-done-advancing` — I'm not sure whether this is the best way. 

(I believe those two tests are failing because of some problem fetching from Docker servers? Not sure)

PR #519, opened 2020-10-16
Fix #518: "Proof using" mode corrupts "Proof with tac".

### LINKED ISSUE #518: In Coq mode, "Proof using" mode corrupts "Proof with tac"
In Coq, with Coq->"Proof using" mode -> ask or always set, and the proof is started with "Proof with tac.", PG will corrupt that Proof line by inserting the "using" before the "with" instead of after, and by deleting the "w" from "with".  One gets things like "Proof using ith tac." instead of the proper "Proof with tac using.".  Because PG allows this mode to alter locked regions, the proof is not undone and the corrupted line can go unnoticed until the file is reparsed.

This is using PG 4.5-git in Emacs 26.1 with Coq 8.12.

Matafou:
Fixing the auto insertion of `Proof using` in the case of a `with` annotation also present.

### DIFF
--- a/coq/coq.el
+++ b/coq/coq.el
@@ -2621,9 +2621,28 @@ Warning: this makes the error messages (and location) wrong.")
   (save-excursion (string-match "\\<as\\>" s)))
 
 ;::::::::::::: inserting suggested Proof using XXX... ;;;;;;;;;;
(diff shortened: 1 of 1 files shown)


Matafou:
rebased.

PR #523, opened 2020-11-01
parallel background compilation redesign without clone jobs

hendriktews:
This is a relative invasive redesign of parallel background compilation for Coq. I tested it a little bit, but I am not doing much Coq nowadays. Volunteers for trying it out on bigger projects are welcome!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -95,3 +95,45 @@ jobs:
           endGroup
 
   # - run: echo "::remove-matcher owner=ert-problem-matcher::"
+
+  compile-test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        coq_version:
+          - '8.9'
+          - '8.10'
+          - '8.11'
+        ocaml_version:
+          - minimal
+      # at most 20 concurrent jobs per free account
+      # cf. https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#usage-limit
+      max-parallel: 6
+      # don't cancel all in-progress jobs if one matrix job fails:
+      fail-fast: false
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Add ert problem matcher
+      run: echo "::add-matcher::.github/ert.json"
+      
+    - uses: coq-community/docker-coq-action@v1
+      id: docker-coq-action
+      with:
+        opam_file: 'dummy.opam'
+        coq_version: ${{ matrix.coq_version }}
+        ocaml_version: ${{ matrix.ocaml_version }}
+        custom_script: |
+          startGroup Print opam config
+            opam config list; opam repo list; opam list
+          endGroup
+          startGroup Install emacs
+            sudo apt-get update -y -q
+            sudo DEBIAN_FRONTEND=noninteractive apt-get install -y -q --no-install-recommends emacs
+          endGroup
+          startGroup Run tests
+            sudo chown -R coq:coq ./ci
+            make -C ci/compile-tests test
+          endGroup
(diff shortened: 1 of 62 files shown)


hendriktews:
I added a bunch of tests that can also be enabled in CI at github. I am therefore quite confident, that the core functionality (compilation without errors) works fine. 
@erikmd @Matafou @cpitclaudel : How to proceed? Shall I simply merge, if I feel confident?

Matafou (commented):
Did you test with tramp?

Matafou:
I think you can merge when you feel confident. And be ready to fix if bugs are detected :-).
Thanks for this job. I recently got the confirmation that this is considered a "killer feature" of pg by the coq team :-).

erikmd:
@hendriktews 

> How to proceed? Shall I simply merge, if I feel confident?

Sure! I fully agree with @Matafou 's comment.

FYI I've just restarted the CI test jobs of this PR, as #529 is now fixed.

erikmd:
BTW @hendriktews 

>I added a bunch of tests that can also be enabled in CI at github

did you already enable these tests from the github action? or do you need any hint to perform this?

erikmd:
Hi @hendriktews, FYI the restart of the action lead to a run on exactly the same commit
(https://github.com/ProofGeneral/PG/commits/587b4539d3c6217d9faa23f9180cb91365197ac4), which does not take into account the last version of master (https://github.com/ProofGeneral/PG/commit/ccb37e8026e208825c10f4d5ce64378430e5f7a3)

So I guess you'll need to push another commit if you want to restart all the CI tests from scratch :)

hendriktews:
> Did you test with tramp?

Do you mean compilation and tramp? This did not work before and
does not work now, because the require commands from the
scripting buffer are processed by coqdep in temporary files.
@maxim ones created PR 409 for coq to make it possible to do this
without temporary files. Unfortunately I dropped the ball back
then and this PR never made it.

hendriktews:
> did you already enable these tests from the github action?

No. They first need to land on master, right?

I currently have 6 tests in 5 subdirectories in
`ci/compile-tests`. I can run them all together, but probably it
makes more sense to register each subdirectory separately in CI.

Matafou:
Le dim. 6 déc. 2020 à 18:08, hendriktews <notifications@github.com> a
écrit :

> > Did you test with tramp?
>
> Do you mean compilation and tramp? This did not work before and
> does not work now, because the require commands from the
> scripting buffer are processed by coqdep in temporary files.
> @maxim ones created PR 409 for coq to make it possible to do this
> without temporary files. Unfortunately I dropped the ball back
> then and this PR never made it.


Nevermind. But I was speaking about your fix of coq-callcoq itself. I was
afraid of a regression there.

P.

erikmd:
@hendriktews 

> did you already enable these tests from the github action?
> No. They first need to land on master, right?

No, as soon as GitHub Actions have been already used, any PR can trigger new tests (so that we can also "test the tests" before they are merged :)

> I currently have 6 tests in 5 subdirectories in `ci/compile-tests`. I can run them all together, but probably it makes more sense to register each subdirectory separately in CI.

It's up to you. Each separate job comes from an item in this file:
https://github.com/ProofGeneral/PG/blob/ccb37e8026e208825c10f4d5ce64378430e5f7a3/.github/workflows/test.yml#L13

There are several possibilities, depending on:
* whether you want to "parallelize" the tests of the various directories (in which case you need a `matrix:` spec in the job, but maybe a "sequential" run of the 6 tests is OK?)
* and whether you need a working Coq at hand (I guess so?), in which case you could just add lines here: https://github.com/ProofGeneral/PG/blob/ccb37e8026e208825c10f4d5ce64378430e5f7a3/.github/workflows/test.yml#L94-L95

hendriktews:
> Nevermind. But I was speaking about your fix of coq-callcoq itself. I was
> afraid of a regression there.

I tested that with ssh to a different machine with a different
coq version.

Matafou:
Great

> > Nevermind. But I was speaking about your fix of coq-callcoq itself. I was afraid of a regression there.

> I tested that with ssh to a different machine with a different coq version.

Great. Thanks again!

PR #526, opened 2020-11-20
fix coq-callcoq for emacs 27

### LINKED ISSUE #525: coq version detection breaks in emacs 27
Hi,

this morning I got emacs 27 with Debian testing and this apparently breaks the automatic coq version detection in PG. AFAICS, it is the call of `shell-command` inside `coq-callcoq` inside `coq-system.el`. In emacs 26 this seems to return the process exit status. However in emacs 27 it seems to return always 1. Just try
```elisp
(with-temp-buffer
  (let ((ret (shell-command "/usr/bin/true" (current-buffer))))
    (cons ret (buffer-string))))
```
`shell-command` is not documented in the elisp manual and its docstring does not say anything about the return value. Therefore I would say we have been relying on unspecified behavior here.
@cpitclaudel : I believe you added those lines in 2016 - what is your opinion?
Hendrik

hendriktews:
Use process-file and omit find-file-name-handler, because
process-file takes care of file handlers already.

Fixes #525

### DIFF
--- a/coq/coq-system.el
+++ b/coq/coq-system.el
@@ -112,21 +112,14 @@ The given option should make coqtop return immediately.
 Optionally check the return code and return nil if the check
 fails.
 This function supports calling coqtop via tramp."
-  (let* ((default-directory
-           (if (file-accessible-directory-p default-directory)
-               default-directory
-             "/"))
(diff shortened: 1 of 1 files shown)


hendriktews:
Works for me with a simple test with trump for emacs 27. Could somebody else try with a different emacs version?

PR #527, opened 2020-12-02
Fix byte compilation

rgrinberg:
load-file-name is not set to pg's source when compiling from doom-emacs

https://github.com/hlissner/doom-emacs/issues/2788

### DIFF
--- a/proof-general.el
+++ b/proof-general.el
@@ -56,8 +56,9 @@
 ;;;###autoload
 (eval-and-compile
   (defvar pg-init--script-full-path
-    (or (and load-in-progress load-file-name)
-        (bound-and-true-p byte-compile-current-file)
+    (or (bound-and-true-p byte-compile-current-file)
+        (and load-in-progress load-file-name)
+
         (buffer-file-name)))
   (defvar pg-init--pg-root
     (file-name-directory pg-init--script-full-path)))


cpitclaudel:
! this is quite confusing.  Can you clarify what's going on? Is it the case that both `load-file-name` and `byte-compile-current-file` are set and yet they disagree?

rgrinberg:
> Is it the case that both load-file-name and byte-compile-current-file are set and yet they disagree?

Exactly. In particular, `load-file-name` is set to doom's upgrade script `doom.el`

cpitclaudel:
OK, I see.  Is this because doom.el is `load`ed and then triggers a byte-compilation?

I want to wait for @monnier 's opinion on this, but then indeed it might be more reasonable to flip these two things.

rgrinberg:
> OK, I see. Is this because doom.el is loaded and then triggers a byte-compilation?

That's right. Dune's package manager is an .el script that runs from the command line. After it updates the package sources, it will byte compile them.

cpitclaudel:
Got it, thanks.

cpitclaudel:
Then I'm OK with this patch; I think we should also make a bug report on the Emacs side to provide a function to return the name of the current file, since I've seen this kind of code with load-file-name and byte-compile-current-file in a bunch of places.

rgrinberg:
Do you think it's OK to merge this workaround for now? This breaks the installation of PG for quite a few users, and I'm under the impression that you don't want these hacks in the long term anyway.

cpitclaudel:
Yeah, works for me :) Thanks for the ping!

PR #532, opened 2020-12-09
Compile tests

hendriktews:
This enables the new test for background compilation in CI at github. I make a separate PR, because I don't expect it to work immediately.


hendriktews:
The tests work now. I continue in PR #523.

PR #536, opened 2020-12-25
make-temp-file without text argument for emacs 25

### LINKED ISSUE #534: Compile Before Require not working anymore
Hello everyone,

I'm using emacs 25.2.2 with proof-general 20201219.1543 (installed through melpa). I'm working on a Coq project, with Coq version 8.12.

When I enable the "Compile Before Require" option in the menu bar and I hit "C-c Enter" to get all my required libraries compiled, then proof-general stalls at the first "Require Import" line.

In the Messages buffer, it says:

> Unexpected error during parallel compilation: (wrong-number-of-arguments (1 . 3) 4)
> proof-extend-queue: Wrong number of arguments: (1 . 3), 4

I don't if it comes from my set up (maybe my Emacs version is too old) or from proof-general.

Thanks,

Vincent

hendriktews:
fixes #534

### DIFF
--- a/coq/coq-par-compile.el
+++ b/coq/coq-par-compile.el
@@ -1639,9 +1639,10 @@ was queued."
            (get job 'load-path)))
         (coq-load-path-include-current nil)
         (require-command
-         (mapconcat 'identity (nth 1 (car (get job 'queueitems))) " ")))
-    (put job 'temp-require-file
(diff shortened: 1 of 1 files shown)


Matafou:
Has the failing test something to do with this PR? I suppose it does not.

PR #552, opened 2021-02-22
protect uses of coq-callcoq

### LINKED ISSUE #551: Recent changes broke PG when coqtop can't be found
With https://github.com/ProofGeneral/PG/pull/526 , my students are getting errors:

```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  string-match("-topfile" nil)
  coq--supports-topfile()
  (let ((topfile-supported (coq--supports-topfile))) (append (if (and topfile-supported buffer-file-name (coq-regular-filename-p buffer-file-name)) (cons "-topfile" (cons buffer-file-name nil)) (if (and topfile-supported buffer-file-name) (message "Warning: this Coq buffer is probably not compilabl...")) nil) (cons "-emacs" (coq-coqc-prog-args loadpath current-directory pre-v85))))
  coq-coqtop-prog-args(nil)
  coq-prog-args()
  eval((coq-prog-args) t)
```

Repro: set `coq-prog-name` to `"AAA"`, then `M-: (coq-prog-args)` will fail because `coq-autodetect-help` will return nil.

cc @hendriktews

hendriktews:
Uses of coq-callcoq need to correctly handle nil as result for
the case that coq-callcoq fails.

Additionally, add a regression test.

Fixes #551

### DIFF
--- a/coq/coq-system.el
+++ b/coq/coq-system.el
@@ -140,7 +140,10 @@ Interactively (with INTERACTIVE-P), show that number."
 (defun coq-autodetect-help (&optional interactive-p)
   "Record the output of coqotp -help in `coq-autodetected-help'."
   (interactive '(t))
-  (setq coq-autodetected-help (coq-callcoq "-help")))
+  (let ((coq-output (coq-callcoq "-help")))
+    (if coq-output
+        (setq coq-autodetected-help coq-output)
+      (setq coq-autodetected-help ""))))
 
 
 (defun coq--version< (v1 v2)
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -131,6 +131,7 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
+      # change to selected versions of coq-emacs docker containers
       matrix:
         coq_version:
           - '8.9'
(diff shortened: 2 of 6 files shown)


coq/coq-system.el
```diff
@@ -140,7 +140,10 @@ Interactively (with INTERACTIVE-P), show that number."
 (defun coq-autodetect-help (&optional interactive-p)
   "Record the output of coqotp -help in `coq-autodetected-help'."
   (interactive '(t))
-  (setq coq-autodetected-help (coq-callcoq "-help")))
+  (let ((coq-output (coq-callcoq "-help")))
+    (if coq-output
+        (setq coq-autodetected-help coq-output)
+      (setq coq-autodetected-help ""))))
```
cpitclaudel, line 146: I think this is the same as ` (setq coq-autodetected-help (or (coq-callcoq "-help") ""))`?
Are we sure there were no other callers that might be affected by the callcoq change?

Matafou: I don't understand the question. `coq-callcoq` is not modified, only this call to it. The only other call (10 lines above) is already protected.

Matafou (approved):
LGTM.

coq/coq-system.el
cpitclaudel: I hadn't checked all the callers.  The bug that this PR fixes was introduced by modifying callcoq, so it seemed natural to make the fix there too.

Matafou: I see. I think this is OK like this.

erikmd (approved):
I agree with @cpitclaudel's [comment](https://github.com/ProofGeneral/PG/pull/552/files#r580650812) that the phrasing of the fix is unnecessary verbose, but LGTM as well.

Then regarding [the other question](https://github.com/ProofGeneral/PG/pull/552/files#r582816426) raised by Clément: the `nil` return value is coherent with the current docstring (below). So, merging this blocker bugfix now :)

```
coq-callcoq is a compiled Lisp function in ‘coq-system.el’.

(coq-callcoq OPTION &optional EXPECTEDRETV)

Call coqtop with the given OPTION and return the output.
The given option should make coqtop return immediately.
Optionally check the return code and return nil if the check
fails.  Return also nil on other kinds of errors (e.g., ‘coq-prog-name’
not found).
This function supports calling coqtop via tramp.
```

coq/coq-system.el
hendriktews: > I hadn't checked all the callers. The bug that this PR fixes was introduced by modifying callcoq, so it seemed natural to make the fix there too.

A simple grep shows there are only 2 uses of `coq-callcoq`. The first one protected itself already against a return value nil for a long time. Further, returning the empty string from `coq-callcoq` when `coq-prog-name` cannot be found --which was the case before PR #526-- is a design error IMO, especially, because there has always been a code path that returned nil. I considered all this before preparing this PR and therefore decided to protect the only remaining use of `coq-callcoq`.

PR #553, opened 2021-02-25
test: Add Emacs 27.1 in CI & Bump the minimal-required version of Emacs (24.5 for now)

erikmd:
This small PR is a follow-up of the discussion I had with @hendriktews last month.

To sum up:

* Although it is nice to support many emacs versions, and periodically add more versions within the PG CI configuration,
  it is also useful to drop support for very old emacsen at some point. Notably to be able to benefit from newest features :slightly_smiling_face:

* A possible policy that looked reasonable for @hendriktews and me is documented in the [`README.md` from this PR](https://github.com/ProofGeneral/PG/blob/3b93f2e1149ebd13e97d6fd4954269c07fe210bd/README.md#installing-proof-general).
  The MELPA/package.el metadata will naturally be updated as well by this PR.

* **Summary:** support Emacsen from the versions packaged in **Debian Stable** and **Ubuntu LTS until EoS**; which implies to:
  * Skip support for Emacs 24.3, 24.4 just now;
  * Skip support for Emacs 24.5 in May 2021 (given it is only packaged in Ubuntu 16.04 LTS, supported [until](https://wiki.ubuntu.com/Releases) April 2021)

Are you OK with this? Cc @ProofGeneral/core ; hopefully I could merge this PR by this Friday if you approve this.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,15 +17,15 @@ jobs:
     strategy:
       matrix:
         emacs_version:
-          - 24.3
-          - 24.4
           - 24.5
+          # 24.5 to be removed at End-of-Support of Ubuntu 16.04 LTS (https://wiki.ubuntu.com/Releases)
           - 25.1
           - 25.2
           - 25.3
           - 26.1
           - 26.2
           - 26.3
+          - 27.1
       # at most 20 concurrent jobs per free account
       # cf. https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#usage-limit
       max-parallel: 4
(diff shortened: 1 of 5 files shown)


Matafou:
I am OK with this.

Matafou:
BTW what newer features would you plan to use?

erikmd:
> BTW what newer features would you plan to use?

Personally, a feature I'd like to use at some point is the [`make-process`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Asynchronous-Processes.html#index-make_002dprocess) function, which is more flexible than `start-process` etc.
(it is available [since Emacs 25.1](https://github.com/emacs-mirror/emacs/blob/emacs-25.1/src/process.c#L1353)).

Apart from that, just to give a small code example which could benefit from this first bump (to 24.5): you have this code committed by @cpitclaudel :)

https://github.com/ProofGeneral/PG/blob/bdb67820696b2bfec06fda72f2a39548c73dcaf7/coq/coq-system.el#L146-L152

erikmd:
Regarding this PR (whose impact is moderate, but has the benefit to document "more formally" the Emacs versions we support),
I propose to merge in ~24h, if no one objects by then.

PR #554, opened 2021-02-25
feat: Add proof-upgrade-menu triggering proof-upgrade-elpa-packages

erikmd:
We sometimes have bug reports that amount to noticing that the version of PG that is installed is not fully up-to-date.

And even if the upgrade procedure is documented in https://proofgeneral.github.io/#keeping-proof-general-up-to-date (assuming a MELPA installation), I reckon it is not very straightforward to perform one such update.

Hence this PR that relies on package.el (and behaves in a safe way if the feature is not available thanks to `proof-try-require`).
To sum up, it is **merely** a shortcut for doing <kbd>M-x package-list-packages RET</kbd>, <kbd>r</kbd>, <kbd>U</kbd>, <kbd>x</kbd>, but forcing the use of a mini-buffer confirmation rather than a dialog-box for better UX, letting the end user to validate or not with `y/n`.

Below is one screenshot to give an overview of the location of the menu item provided by this PR.

2021-02-25_proof-upgrade-elpa-packages

Cc @ProofGeneral/core are you OK for merging this?

If yes, I could then update https://proofgeneral.github.io/#keeping-proof-general-up-to-date doc to also suggest this simple way :)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -8,6 +8,11 @@ and the PG Trac http://proofgeneral.inf.ed.ac.uk/trac
 
 ** Generic changes
 
+*** new command and menu item to easily upgrade all packages
+    - To upgrade all ELPA packages (including ProofGeneral if it was
+      installed via MELPA), do "M-x proof-upgrade-elpa-packages RET"
+      or use the "Proof-General > Upgrade ELPA packages..." menu item
+
 *** bug fixes
     - Using query-replace (or replace-string) in the processed region
       doesn't wrongly jump to the first match anymore.
--- a/generic/proof-autoloads.el
+++ b/generic/proof-autoloads.el
@@ -17,8 +17,8 @@
 (provide 'proof-autoloads)
 
 
-;;;### (autoloads nil "../coq/coq" "../coq/coq.el" (23587 63448 184007
-;;;;;;  512000))
+;;;### (autoloads nil "../coq/coq" "../coq/coq.el" (24631 49612 444031
+;;;;;;  796000))
 ;;; Generated autoloads from ../coq/coq.el
 
 (autoload 'coq-pg-setup "../coq/coq" "\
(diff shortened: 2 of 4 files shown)


Matafou:
I did not test it but I think this is a very good idea. Adding a line in the CHANGES file with enough explanations (menu description + command name for menu haters) may be a good idea. I don't know what is best for advertising this: doc vs menu vs CHANGES.

cpitclaudel:
:+1: Great work! Are you sure you want the button to do upgrade all, rather than upgrading just PG and its dependencies?

erikmd:
Hi, thanks for your feedback! So:
- Good point regarding the CHANGES file, I will push a "doc commit". Actually regarding the "doc vs. menu vs. CHANGES", I'd say the best strategy is probably to replace `vs.` with `and` ;) so this should be documented in the https://proofgeneral.github.io/ page as well
- Indeed Clément, upgrading all package is on-purpose… but maybe we could add a prefix argument, so that:  
  *default* action = upgrade all packages, and
  alternative action = upgrade only PG / company-coq and their deps… but I don't know if this is feasible easily.

So don't you mind Clément if I merge this PR after updating the CHANGES files, and postpone the enhancement w.r.t. the prefix arg?

erikmd:
> I will push a "doc commit"

Done − I've also turned the `proof-upgrade-elpa-packages` function into an autoload BTW.

erikmd:
> don't you mind Clément if I merge this PR after updating the CHANGES files, and postpone the enhancement w.r.t. the prefix arg to another PR?

So the PR looks ready now; @cpitclaudel no objection for merging? (sorry to ping you again :)

cpitclaudel:
Go for it :)

erikmd:
Just FYI I've just added a screenshot in https://proofgeneral.github.io/#keeping-proof-general-up-to-date to document this

(as well as the underlying command, following Pierre's suggestion − `M-x proof-upgrade-elpa-packages RET`)

PR #559, opened 2021-03-16
add feature to omit complete opaque proofs

hendriktews:
This commit adds a feature to recognize complete opaque proofs in
the asserted region and to replace them with an admitted proof.
This can drastically improve the processing time for the asserted
region at the cost of not checking the omitted proofs. Omitted
proofs are displayed in some shade of purple in the locked
region.

With this commit, the feature is supported for Coq for files in
which proofs are started with some form of Proof using and
ended with either Qed, Defined or Admitted.

To enable, configure proof-omit-proofs-option or click
Proof General -> Quick Options -> Processing -> Omit Proofs.

### DIFF
--- a/doc/ProofGeneral.texi
+++ b/doc/ProofGeneral.texi
@@ -1495,6 +1495,42 @@ will stop exactly where the proof script fails, showing you the error
 message and the last processed command.  So you can easily continue
 development from exactly the right place in the script.
 
+In normal development, one often jumps into the middle or to the
+end of some file, because this is the point, where a lemma must
+be added or a definition must be fixed. Before starting the real
+work, one needs to assert the file up to that point, usually with
+@kbd{C-c C-RET} (@code{proof-goto-point}). Even for medium sized
+files, asserting a big portion can take several seconds. There
+are different ways to speed this process up.
+@itemize @bullet
+@item
+One can split the development into smaller files. This works
+quite well with Coq, automatic background compilation,
+@ref{Automatic Compilation in Detail}, and the fast compilation
+options, @ref{Quick and inconsistent compilation}.
+@item
+One can configure @code{proof-omit-proofs-option} to @code{t} to
+omit complete opaque proofs when larger chunks are asserted. A
+proof is opaque, if its proof script or proof term cannot
+influence the following code. In Coq, opaque proofs are finished
+with @code{Qed}, non-opaque ones with @code{Defined}. When this
+omit proofs feature is configured, complete opaque proofs are
+silently replace with a suitable cheating command
+(@code{Admitted} for Coq) before sending the proof to the proof
+assistant. For files with big proofs this can bring down the
+processing time to 10% with the obvious disadvantage that errors
+in the omitted proofs go unnoticed. For checking the proofs
+occasionally, a prefix argument for @code{proof-goto-point} and
+@code{proof-process-buffer} causes these commands to disregard
+the setting of @code{proof-omit-proofs-option}. Currently, the
+omit proofs feature is only supported for Coq.
+@item
+An often used poor man's solution is to collect all new material
+at the end of one file, regardless where the material really
+belongs. When the final theorem has been proved, one cleans up
+the mess and moves all stuff where it really belongs.
+@end itemize
+
 Here is the full set of script processing commands.
 
 @c TEXI DOCSTRING MAGIC: proof-assert-next-command-interactive
@@ -3640,6 +3676,21 @@ If non-nil, format for newlines after each command in a script.
 The default value is @code{t}.
 @end defopt
 
+@c TEXI DOCSTRING MAGIC: proof-omit-proofs-option
+@defvar proof-omit-proofs-option 
+Set to t to omit complete opaque proofs for speed reasons.@*
+When t, complete opaque proofs in the asserted region are not
+sent to the proof assistant (and thus not checked). For files
+with big proofs this can drastically reduce the processing time
+for the asserted region at the cost of not checking the proofs.
+For partial and non-opaque proofs in the asserted region all
+proof commands are sent to the proof assistant.
+
+Using a prefix argument for @samp{@code{proof-goto-point}} (M-x @code{proof-goto-point})
+or @samp{@code{proof-process-buffer}} (M-x @code{proof-process-buffer}) temporarily
+disables omitting proofs.
+@end defvar
+
 @c TEXI DOCSTRING MAGIC: proof-prog-name-ask
 @defopt proof-prog-name-ask 
 If non-nil, query user which program to run for the inferior process.
@@ -4293,6 +4344,7 @@ assistant.  It supports most of the generic features of Proof General.
 * Using the Coq project file::
 * Proof using annotations::
 * Multiple File Support::
+* Omitting proofs for speed::
 * Editing multiple proofs::
 * User-loaded tactics::
 * Indentation tweaking::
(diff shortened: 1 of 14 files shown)


hendriktews:
For files with big proofs I measure a reduction of the processing time of the asserted region to 10%. For files with small proofs and many definitions it goes down to 30-40%.
Documentation in the manual, a test and a prefix argument to selectively process all proofs will follow in separate commits.

erikmd:
Hi @hendriktews, thanks, that looks very nice! :)

And let us know when/if you'd like we test the feature ^^

hendriktews:
This new test `omit-proofs-omit-and-not-omit` fails now and then because of #563.

hendriktews:
> And let us know when/if you'd like we test the feature ^^

It's ready for testing now. All you need to know is documented in `proof-omit-proofs-option`.

hendriktews:
I consider this now complete. I have used the feature for a while for some 2000 lines file and it reduced waiting time there considerably. Does anybody want to try?

erikmd:
> I consider this now complete. I have used the feature for a while for some 2000 lines file and it reduced waiting time there considerably. Does anybody want to try?

Hi @hendriktews, thanks, I'd like to try − as soon as I can (not this week but before our next meeting on 14th April of course), anyway feel free to merge if others people @ProofGeneral/core were already able to take a closer look :)

Thanks a lot :+1:

Matafou:
I definitely want to try.

hendriktews:
changed according to discussion in maintainer telco: Support proof starts without using and treat Abort as end of a non-opaque proof. @erikmd Do you want to check whether it works now in your file that contains Abort?

erikmd (approved):
Thanks @hendriktews :+1: yes I've just fetched your last commits and tested on [that example file](https://github.com/validsdp/coq-floats-jfla2021/blob/master/demo_coq_floats.v), so I can confim it works very fine!

erikmd:
However there's a small issue I guess: the prefix for skipping the proof-omitting feature does not work with Company-Coq.

erikmd:
@hendriktews to be more precise: after installing the `company-coq` package from MELPA and enabling it, `M-x company-coq-mode RET`, I get: <kbd>C-h k C-c C-RET</kbd> ↓

```
C-c <C-return> runs the command company-coq-proof-goto-point (found in
company-coq--core-map), which is an interactive compiled Lisp function
in ‘company-coq.el’.

It is bound to C-c RET, C-c <C-return>, <menu-bar> <Proof-General>
<Goto Point>.

(company-coq-proof-goto-point &rest ARGS)

Pass ARGS to ‘proof-goto-point’, hiding company dialog.
```

Do you think it's feasible to overcome this limitation directly from PG? or maybe we can just open a PR in [company-coq's repo](https://github.com/cpitclaudel/company-coq).

doc/ProofGeneral.texi
```diff
@@ -5135,6 +5191,41 @@ General will perform some unnecessary compilations.
...
+To speed up asserting larger chunks, Proof General can omit
+complete opaque proofs by silently replacing the whole proof
+script with @code{Admitted}, @ref{Script processing commands}.
+This works when
+@itemize
+@item
+proofs are not nested
+@item
+opaque and non-opaque proofs start with @code{Proof.} or
+@code{Proof using}
```
erikmd, line 5206: @hendriktews BTW I believe it could be worth it to add in the manual, a comment/disclaimer/advice saying that:

> If the proof script relies on sections, it is highly recommended to use a `Proof using` clause for each lemmas contained in the sections, so that the Admitted lemmas are generalized in an accurate way. For completeness, note that PG provides a dedicated feature to generate these `Proof using` clauses automatically (a defective form being e.g. `Proof using Type` if no hypothesis is used): see the menu command `Coq > "Proof using" mode`…

(but feel free to amend the prose!)

hendriktews:
company-coq binds a different command to `C-c C-return`, then it is not really a surprise, that it is not working, is it?

> Do you think it's feasible to overcome this limitation directly from PG? or maybe we can just open a PR in [company-coq's repo](https://github.com/cpitclaudel/company-coq).

No, `company-coq-proof-goto-point` needs a different interactive declaration. Try to change that into `(interactive "P")`. With `C-c C-b` (`proof-process-buffer`) the prefix works also in company-coq?

erikmd:
@hendriktews 

> company-coq binds a different command to `C-c C-return`, then it is not really a surprise, that it is not working, is it?

Exactly, I just wanted to mean this :) that the issue was related to the rebinding.

> No, `company-coq-proof-goto-point` needs a different interactive declaration. Try to change that into `(interactive "P")`.

Yes that seems to suffice. So I opened PR https://github.com/cpitclaudel/company-coq/pull/247

> With `C-c C-b` (`proof-process-buffer`) the prefix works also in company-coq?

Yes!

BTW, do you agree with the suggestion I had mentioned to extend the manual w.r.t. "Proof using" ?

Anyway when `C-u C-c C-RET` will also work with Company-Coq, it'll be awesome! this PR is definitely a nice feature :)

cpitclaudel:
:+1: :clap: great job :)

doc/ProofGeneral.texi
hendriktews: There is already Section 11.3 Proof using annotations, which recommends `Proof using`. Moreover, a `Proof using` recommendation has not much to do with this PR. May I therefore suggest that you check Section 11.3 and amend it, if you believe it is not sufficient?

hendriktews:
> BTW, do you agree with the suggestion I had mentioned to extend
> the manual w.r.t. "Proof using" ?

See my comment there (which I entered already the other day, but
github decided to discard it, maybe because I clicked resolve
before comment.

cpitclaudel:
Can we merge this? This fix  to company-coq isn't backwards-compatible.

erikmd:
Hi @hendriktews! sorry for late reply.

> Moreover, a `Proof using` recommendation has not much to do with this PR.

I'm not fully convinced that this PR has not much to do with `Proof using` :sweat_smile: 

Indeed, if there is no `Proof using` annotation, the feature replacing `Qed` with `Admitted` will often make the script fail!

Here is a toy example:

```coq
(* Proof-General > Quick Options > Omit Proofs *)
Section Foo.
  Variables (A : Type) (other : True).
  Lemma id : A -> A.
  Proof .
    auto.
  Qed.
End Foo.
Fail Definition bar := id nat 0. (* while it would be OK with `Proof using A.` above *)
```

Hence [my earlier suggestion](https://github.com/ProofGeneral/PG/pull/559#discussion_r615158228) to add a small note in 11.5, advising to read 11.3 as well :)

But if you strongly disagree with this addition, maybe @Matafou (which implemented the `Proof using` helper feature IINM and also discussed this topic at the telco) could help to come to a decision?

Matafou:
`Proof using` has definitely to do with this PR. If you use this in a section that is not decorated with `proof using` then after section ending coqtop will compute a wrong type for each lemma of the section. It should be made clear in the doc and in the CHANGE.

erikmd (approved):
Thanks @Matafou for your comment!
So I'm going to readily push a commit that documents that Proof-using requirement in the manual and file CHANGES.
Apart from this detail, the PR looks very fine, so @Matafou / @hendriktews / @cpitclaudel, feel free to merge when you see fit (otherwise I'll do it on tomorrow Wednesday in the evening).

Thanks again Hendrik, for this very nice feature!

erikmd:
OK, so let's merge this now.

hendriktews:
Thanks for polishing and merging. The argument about sections and proof using is certainly valid and it is good to have this in the manual (although I am not sure if anybody ever reads the manual).

PR #569, opened 2021-04-07
enable Tramp support for Emacs 26 or later

hendriktews:
tramp support is not complete with these patches, but works if used with care. There are a few assertions that quick user interactions can trigger. Automatic background compilation via tramp is very slow, because tramp opens a new connection with all overhead for each asynchronous coqdep and coqc invocation. With these patches, tramp support only works for Emacs version 26 or later, because it uses make-nearby-temp-file.


Matafou:
FTR:

- slowlyness was solved by @hendriktews (reusing tramp connexion)
- the bug which triggers some assertion is still there as far as we can tell.

Merging now would need to warn users about not triggering any tramp feature during background compilation.

PR #577, opened 2021-05-12
coq-par-compile: use doubly graph to unlock ancestors efficiently

### LINKED ISSUE #572: ProofGeneral freezes when there is a build error in a dependency file
With "build dependencies" mode enabled, usually PG shows error messages when a dependency fails to build. Recently, it started to just hang instead, so that I have to kill emacs entirely.

My Coq/PG-related settings are:
```
 '(coq-compile-before-require t)
 '(coq-compile-parallel-in-background t)
 '(coq-compile-quick 'no-quick)
 '(coq-compile-vos 'vos)
 '(coq-one-command-per-line nil)
 '(proof-follow-mode 'followdown)
 '(proof-next-command-insert-space nil)
 '(proof-splash-enable nil)
 '(proof-three-window-enable t)
```

hendriktews:
Change from a singly linked graph (only childs point to parents)
to a doubly linked graph (also parents point to childs). Instead
of propagating ancestors upwards as compilation progresses,
ancestors are now collected when require jobs are retired. This
makes it easier to avoid exponential blowup when traversing the
graph. Additionally, completely link ancestors that are ready
into the graph. Fixes #572.

### DIFF
--- a/ci/compile-tests/001-mini-project/runtest.el
+++ b/ci/compile-tests/001-mini-project/runtest.el
@@ -36,6 +36,7 @@
 (ert-deftest cct-mini-project ()
   "Test successful background compilation and ancestor recording."
   (find-file "a.v")
+  ; (setq coq--debug-auto-compilation t)
   (cct-process-to-line 25)
   
   (cct-check-locked 24 'locked)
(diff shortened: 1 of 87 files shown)


RalfJung:
I can confirm that with this PR, I cannot reproduce https://github.com/ProofGeneral/PG/issues/572 any more.

Issue #568, opened 2021-03-30
Why does PG show stale goals?


cpitclaudel:
I don't know what happened of late, but over the last few years it's become more and more common that I have to ask PG to refresh the goal explicitly.  Often the goal window stays blank, or it shows an outdated goal.  If a comment follows a bit of code and both are processed together, PG doesn't show the goal. And so on.

What's going on?  It's come to the point that C-c C-p is one of the first PG keybindings I teach my students.  Is it our fault, or is it on Coq's side?

hendriktews:
I would say it is the silent processing together with the difficulty to reliably insert something in proof-action-list. I believe it is our fault. With 8d545bca and 14ea1a3f it will become easier to insert stuff into proof-action-list.

cpitclaudel:
Thanks.  Then I think we should treat your work as critical-priority.  At the moment, PG is basically unusable for newcomers.  Teaching a Coq class is a real reality check on that front.

Matafou:
It is our fault for sure, but it also comes from coq slowly changing its behaviour. I think it spontaneously prints the goal less often. We definitely need to refactor the "Show" stuff.

Matafou:
@hendriktews is it really a problem of adding to the action list? It looks to me that we need a way to issue a "Goal" precisely when there is *nothing left* in the action list (and there is an open proof). Without cleaning the response buffer.

cpitclaudel:
Thanks @Matafou.

hendriktews:
I opened a PR with tests where PG does not show any goal or the wrong one. They cover the cases of
* Proof
* comment at the end of the queue region
* auto
* error in the proof script
If you know another way to get no or the wrong goal in PG, please add a test or tell here about it.

ejgallego:
This is likely due `coqtop` not printing a goal again if it didn't change. For example, after `Proof.`.

Only reliable fix is to actually move to a model where PG queries for goals, and stops playing weird games with Coq output [ PG will always lose]

Issue #103, opened 2016-08-17
Proof state not refreshed after Admitted


tchajed:
In the following example,

```coq
Goal True.
Admitted.
```

If you process the sentences one at a time, the proof state remains at 1 subgoal of True. It seems Admitted is being treated the same as Qed, which is to do nothing. This isn't quite correct, though - for Admitted clearly the proof state has changed, and in a nested proof the `*goals*` window should be updated to show the outer proof again even for a Qed.

psteckler:
Just to note, in my XML protocol fork of PG, for this example, the goal
window is updated after the Admitted has been processed.

-- Paul

Blaisorblade:
FWIW, not refreshing after `Qed.` is also incorrect with nested proofs (this might be the real issue underlying #154)—it should update the screen to show the open goals. If you *only* execute `Qed.`, refresh is only needed if there are nested proofs; but if you jump to after `Qed.` from a different state, the shown goal might even have been solved.

Is that also handled correctly in the XML fork?

psteckler:
The PG/xml fork always updates the goal buffer after each individual
statement, if single-stepping, or after a block of statements, if asserting
to a point or running the whole buffer.

-- Paul

PR #585, opened 2021-06-06
add tests for checking that goals are correctly shown

hendriktews:
Most of these tests currently fail because of different instances
of #568, see also #429, #467.


hendriktews:
All the currently failing tests have `:expected-result :failed`, therefore the checks are reported as passed.

Columbus240:
Maybe also consider #103. Another example similar to #103 : `Goal True. Abort.`

Scutua (approved):

Issue #589, opened 2021-06-25
PG mangles strings in _CoqProject


Columbus240:
In my _CoqProject I have the line
```
-arg -set -arg "'Default Goal Selector=!'"
```
because this way I managed to convince the other Coq tools (coq_makefile, coqc, etc) that they should `Set Default Goal Selector "!".`
But PG produces the following value in `coq-prog-args`: `("-emacs" "-set" "'Default" "Goal" "Selectors=!'")` while I expected `("-emacs" "-set" "Default Goal Selector=!")`.

Columbus240:
I just noticed that this issue is related to #392.

cpitclaudel:
This coq_makefile stuff has been a plague for years.  Hopefully at some point in the future we can get a simple coq project format that's declarative (instead of a bunch of shell commands); something like JSON or TOML or CFG or …

Of course this works with Emacs:

```
-arg -set -arg "\"Default Goal Selector=!\""
```

… but not with _CoqProject...

Thanks for the PR by the way.  I'd be fine merging it as-is; I don't expect it would cause much trouble, and I don't want to spend much time on this exciting source of broken-ness.

Columbus240:
> Of course this works with Emacs:
> 
> ```
> -arg -set -arg "\"Default Goal Selector=!\""
> ```
> 
> … but not with _CoqProject... 

Jup, have noticed that as well. Erm, don’t merge the branch for now. I understand now, why exactly the interaction of coq_makefile, make and coqc is so messed up regarding quotations. I’ll submit a PR to coq in a few hours. If that PR doesn’t get accepted, I’d propose merging my quick fix.

Columbus240:
Another solution that works with coq_makefile but not with ProofGeneral: `-arg "-set 'Default Goal Selector=!'"`. I think this syntax would be acceptable in the current state of affairs (if it were sufficiently well documented).

To simulate the behaviour of the interaction of coq_makefile and make, ProofGeneral would need to do the following:
* Split the input stream into tokens (i.e. a list of strings) as done here: https://github.com/coq/coq/blob/7c91b3cdfa1a7edb27a184a37c481ca61c6a2eaa/lib/coqProject_file.ml#L91-L139.
* Treat all the other options as before.
* Create an empty buffer "ExtraArgs". For every "-arg" followed by a token x, append x to ExtraArgs (with spaces inbetween).
[...]

PR #590, opened 2021-06-25
Handle single-quotes in arguments to `-arg`

Columbus240:
A haphazard sketch to issue #589. It is *not* an acceptable final solution, I think

coq_makefile parses _CoqProject files in a slightly weird way. If we write `-arg "a b"` in _CoqProject then it passes the arguments `a` and `b` to coqc, but if we write `-arg "'a b'"` it passes `a b` to coqc, as a single argument.

I tried to adapt PG to this behavior. It works for my use-case, but I haven’t tested where the differences in behaviour (between PG and coq_makefile) lie now. A weird fact is, that coq_makefile produces some string, which it stores in `Makefile.coq.conf`, which later gets processed by make and passed on to coqc. So a "good" solution would try to reproduce the complete processing chain, i.e. first the processing done by coq_makefile, then the processing done by make.

Maybe we could call [`lib/coqProject_file.mli`](https://github.com/coq/coq/blob/master/lib/coqProject_file.mli) from elisp and hand over the parsing to the OCaml code of the coq repo.

### DIFF
--- a/coq/coq-system.el
+++ b/coq/coq-system.el
@@ -123,7 +123,6 @@ This function supports calling coqtop via tramp."
           (if (or (not expectedretv) (equal retv expectedretv))
               (buffer-string)))
       (error nil))))
-        
 
 (defun coq-autodetect-version (&optional interactive-p)
   "Detect and record the version of Coq currently in use.
@@ -554,16 +553,18 @@ Returns a mixed list of option-value pairs and strings."
 OPTIONS is a list or conses (switch . argument) and strings.
 Note that the semantics of the -arg flags in coq project files
 are weird: -arg \"a b\" means pass a and b, separately, to
-coqtop."
+coqtop. But -arg \"'a b'\" means to pass a and b together."
   (let ((args nil))
     (dolist (opt options)
       (pcase opt
         ((or "-byte" "-op")
          (push opt args))
         (`("-arg" ,concatenated-args)
-         (setq args
-               (append args
-                       (split-string-and-unquote concatenated-args coq--project-file-separator))))))
+	 (if (and (string-prefix-p "'" concatenated-args) (string-suffix-p "'" concatenated-args))
+	     (setq args (append args (list (substring concatenated-args 1 -1))))
+           (setq args
+		 (append args
+			 (split-string-and-unquote concatenated-args coq--project-file-separator)))))))
     (cons "-emacs" args)))
 
 (defun coq--extract-load-path-1 (option base-directory)


Columbus240:
Somehow this seems to work in the cases I want to use. But I’m pretty sure it doesn’t mirror the behaviour of coq_makefile & make in all cases.
If such cases arise in practice, the algorithm can be changed accordingly.

Matafou:
Hi, this seems like a good patch, my main concern is about already written _CoqProject files that would break...
I have the feeling that neither `-arg "a b"` not `-arg "'a b'"` is used very much.
I also have the feeling that this patch reduces differences between pg and coqc so even if both do not match perfectly it still makes the situation better than it is now.
Relying on ocaml code seems too complicated.
One day dune will save us all... Or not.
@cpitclaudel @erikmd do you have an opinion?

Columbus240:
Hi. Since I recently changed how coq_makefile interprets the _CoqProject, we might adapt the algorithm, so it precisely mirrors the behavior of coq.

Matafou:
Hi @Columbus240 can you be more precise please? What is the new behaviour? In what versions of coq(_makefile) can we expect it to be available?

Matafou:
Pinging @Columbus240 you said you had changed the behaviour of `coq_makefile`, can you tell us more about this change please?

erikmd:
> Hi, this seems like a good patch, my main concern is about already written _CoqProject files that would break...

Yes, exactly.

To this aim, I'd just suggest browsing the repositories of Coq libraries that are part of [Coq's CI](https://gitlab.com/coq/coq/-/pipelines/372638339)
→ cf. https://github.com/coq/coq/blob/master/Makefile.ci#L114
→ and https://github.com/coq/coq/tree/master/dev/ci

And among these libraries, a number of them have some `-arg`-based `_CoqProject`, e.g.:
https://github.com/coq/coq/tree/master/dev/ci https://github.com/JasonGross/fiat-crypto/blob/master/_CoqProject.in#L2

> I have the feeling that neither `-arg "a b"` not `-arg "'a b'"` is used very much.
> I also have the feeling that this patch reduces differences between pg and coqc so even if both do not match perfectly it still makes the situation better than it is now.

I didn't take a close look but I believe you're right; if the fix is "conservative" w.r.t. most libraries I suggested above, +1 for merging indeed.

Matafou:
I am still waiting for @Columbus240 to clarify these sentence:

> Hi. Since I recently changed how coq_makefile interprets the _CoqProject, we might adapt the algorithm, so it precisely mirrors the behavior of coq.

If coq_makefile changed its behaviour we should try to understand how it works now before merging this I guess.

Columbus240:
Sorry for letting you wait.
The new behaviour is documented here: https://github.com/coq/coq/blob/master/doc/sphinx/practical-tools/utilities.rst. In particular the paragraph [The Grammar of _CoqProject](https://github.com/coq/coq/blob/master/doc/sphinx/practical-tools/utilities.rst#the-grammar-of-_coqproject) should contain the necessary information.
A very meandering text is also available in the corresponding PR, coq/coq#14558.

The parsing algorithm is implemented in [lib/coqProject_file.ml](https://github.com/coq/coq/blob/master/lib/coqProject_file.ml) if you've got the time to decipher code.

Regarding availability: the PR with the new behaviour has been merged in July but there has not been a release since then. It is thus only available in development versions of Coq. The first version in which the new behaviour will be available is v8.14.

I'm not sure, if the patch I made in this PR is still compatible with the new behaviour for simple cases. I didn't get around to checking it. Certainly, this PR does not mirror the old nor the new behaviour of coq_makefile. If I remember correctly, the parsing code of ProofGeneral processes some escape sequences (it's the call to `split-string-and-unquote` I think). This is something coq_makefile currently does certainly not do.

To give a short overview of the parsing algorithm:
1. Tokenization. Read the file byte by byte and produce a list of strings. Explained in the documentation linked above.
2. Process the list of strings as if they were command line arguments. See [process_cmd_line](https://github.com/coq/coq/blob/266eef24915ce916412a742f935811846e843a07/lib/coqProject_file.ml#L185-L282). Note that a string following "-arg" is further processed (see [line 242](https://github.com/coq/coq/blob/266eef24915ce916412a742f935811846e843a07/lib/coqProject_file.ml#L242)) by [process_extra_args](https://github.com/coq/coq/blob/266eef24915ce916412a742f935811846e843a07/lib/coqProject_file.ml#L158-L183).

I made an exemplary _CoqProject file and went through the processing steps. I hope it's useful and that I didn't make a mistake.
```
lorem "ipsum dolor" "sit'amet" #consectetur "adipiscing \n elit
-arg sed -arg do -arg "eiusmod tempor"
-arg 'incididunt ut' -arg "labore 'et dolore' magna"
-arg "'aliqua ut'"
-R "enim ad" -R "'minim veniam'"
"quis # nostrud"
```
Tokenization produces the following list of strings (where a newline indicates the end of a string)
```
lorem
ipsum dolor
sit'amet
-arg
sed
-arg
do
-arg
eiusmod tempor
-arg
'incidunt
ut'
-arg
labore 'et dolore' magna
-arg
'aliqua ut'
-R
enim ad
-R
'minim veniam'
quis # nostrud
```
Of course the next step wouldn't be able to interpret the arbitrary strings as filenames or arguments. But lets ignore this and consider what lists the `-arg` and the `-R` arguments collected. I.e. how coqc would get called.
The argument list by `-arg` (the one stored in `COQMF_OTHERFLAGS` in the makefile) will look like this: (the `ut'` is removed, because it doesn't count as "belonging to `-arg`" anymore. Processing `-arg` only looks at the immediately following token)
```
sed
do
eiusmod
tempor
incidunt
labore
et dolore
magna
aliqua ut
```
While the argument list concerning `-R` will look like this. I wanted to illustrate here, that the `'` aren't specially interpreted here.
```
-R
enim ad
-R
'minim veniam'
```

Please ask further questions if something is unclear.

erikmd:
Thanks a lot @Columbus240 for your detailed answer! (and sorry for later reply as well)

As already said [by **@cpitclaudel**](https://github.com/ProofGeneral/PG/issues/589#issuecomment-868578902), the inherent discrepancy between both _CoqProject parsers (either elisp or (ocaml+make)-based) has always been an issue; anyway your patch looks very sensible and as @Matafou I agree we should merge this soon,

and as we have a planned PG team telco on this Wednesday evening, we'll be able to discuss whether it can be merged at once… or maybe after adding some unit tests 🙂 (to check that coq_makefile and PG's parsing behave the same on, at least, a few specific cases; relying on PG's CI that can test PG against several different Coq versions…)

BTW (just to be sure, I just skimmed https://github.com/coq/coq/pull/14558/commits), is there another recent change in coq's _CoqProject parser algorithm that should be taken into account for PG? (beyond this support of single-quotes in `-arg` arguments)
→ in this case, we could just as well add some conditional behavior in PG to support both Coq 8.14+ and Coq ≤ 8.13.

Thanks again!

Matafou:
From reading https://github.com/coq/coq/blob/master/doc/sphinx/practical-tools/utilities.rst I think that the only thing to change is the way double and simple quotes are interpreted.
@Columbus240 the documentation implies that spaces and newlines are treated *exactly* the same. Am I right?

erikmd:
Hi all,

> * [ ] add tests?

Indeed @Matafou I guess that adding unit or integrations tests in this PR would be very nice (all the more as we can test from now on _CoqProject + PG with Coq 8.14.* since there is an image `coqorg/coq:8.14`).

These tests could for example, be inspired by the links I gave in [this old comment of mine](https://github.com/ProofGeneral/PG/pull/590#issuecomment-922135852).

> * [ ] leave full compliance with coq_makefile new semantic (in particular wrt newlines) for another PR?
> * [ ] add a coq version testing (>=8.14?)?

However I believe that these two items can wait for another PR: all the more as if/when we have _CoqProject-dedicated tests,
it will be easy to detect if the semantics is broken for a given version of Coq (Coq 8.14 or Coq 8.15), and add conditional elisp code in this case.

@Matafou @Columbus240 if you agree with my remarks above,
I volunteer to help adding one or two commits dedicated to this CI tests in the PR, although I'm afraid this task won't be done before the October-25th week I think (unless if @hendriktews could and would like to have a look meanwhile)

Matafou:
This is bit-rotting. Should we merge this immediately?

Matafou:
This was waiting for much too long. I don't see much risk.

PR #591, opened 2021-07-13
New indentation engine

Matafou:
This PR proposes a new simplified indentation implementation.

It still uses smie. The main design choice for this new version is that the first word of a command is now a token "Com start". Together with the "." token for ending a command this makes the smie grammar parsing much more "local".
    
Users may experience a few changes in the indentation: a few special cases where the old indentation would avoid shifting to the right have been removed.

Non regression testing have been added. But:
    
- [ ] Should the test go in the `ci` directory as I did, or should it go elsewhere? Ideally it would be nice (but not mandatory) to include this tests in the ci @erikmd


monnier:
> It still uses smie. The main design choice for this new version is
> that the first word of a command is now a token &quot;Com
> start&quot;. Together with the &quot;.&quot; token for ending
> a command this makes the smie grammar parsing much more
> &quot;local&quot;.

    ;; Com1 .(1)
    ;; Definition foo := bar .(2)
    ;; ...
    ;; could easily lead to the grammar tree
    ;;
    ;;                        __.(2)
    ;;                    .(1)      \
    ;;                  /      \     ...
    ;;                 /        \
    ;;             Com1        :=
    ;;                       /     \
    ;;                  Def foo    bar
    ;;
    ;; which means that "(1)" is the parent of ":=", which is the parent
    ;; of "Def foo". This leads to hell. We need Def to be the parent. But
    ;; there are hundreds of commands in coq. so we adopt the policy below.
    ;;
    ;; 2. the lexer tries to always treat the first word of a command as a
    ;; dstinguished token "Com start". And a command is of the form ("Com
    ;; start" exp) in smie grammar. This way the top-node of a command is
    ;; always this token.
    ;;
    ;; This way the above script is parsed as:
    ;;
    ;;                        __.(2)
    ;;                    .(1)      \
    ;;                  /      \     ...
    ;;                 /        \
    ;;             Com1     Com start
    ;;                          |
    ;;                         :=
    ;;                       /    \
    ;;                     foo    bar

I haven't looked at the actual code, but the above sounds great to me.


        Stefan

cpitclaudel:
Awesome work.  Thanks Pierre.  I'm not qualified to look at the actual code, but the motivation looks 👍.

A few examples / tests that were developed to indent "correctly" in the current system.  How are the following indented with the new code? (what's below is how they are currently indented)

```
  Instance spec_of_vect_memcpy : spec_of "vect_memcpy" :=
    fnspec! "vect_memcpy" (len: word) (a1_ptr a2_ptr : address) /
          {n1} (a1: VectorArray.t word n1)
          {n2} (a2: VectorArray.t word n2)
          (pr1: word.unsigned len < Z.of_nat n1)
          (pr2: word.unsigned len < Z.of_nat n2)
          R,
    { requires tr mem :=
        (vectorarray_value AccessWord a1_ptr a1 ⋆
                           vectorarray_value AccessWord a2_ptr a2 ⋆ R) mem;
      ensures tr' mem' :=
        tr' = tr /\
        let res := vect_memcpy len a1 a2 pr1 pr2 in
        (vectorarray_value AccessWord a1_ptr (fst res) ⋆
                           vectorarray_value AccessWord a2_ptr (snd res) ⋆ R) mem' }.
```

```
  Instance spec_of_crc32 : spec_of "crc32" :=
    fnspec! "crc32" data_ptr len / (data : list byte) R ~> r,
    { requires tr mem :=
        (word.unsigned len = Z.of_nat (length data) /\
         (listarray_value AccessByte data_ptr data * R)%sep mem);
      ensures tr' mem' :=
        tr' = tr /\
        r = crc32' data /\
        (listarray_value AccessByte  data_ptr data * R)%sep mem' }.
```

```
        <{ Trace := tr;
           Memory := mem;
           Locals := locals;
           Functions := functions }>
        cmd.seq
          (cmd.while
             (expr.op (if signed then bopname.lts else bopname.ltu)
                      (expr.var from_var)
                      (expr.var to_var))
             body_impl)
          k_impl
        <{ pred (nlet_eq vars v k) }>
```

Matafou:
@cpitclaudel there is problem with the <{ }> thing. I will try to fix it before merging.

@monnier I think that tweaking the lexer to parenthesize each command is a good way to go indeed. I should perhaps re-do the lexer(s) completely but for now it seems to work OK.

Matafou:
>         <{ Trace := tr;
>            Memory := mem;
>            Locals := locals;
>            Functions := functions }>
>           cmd.seq
>             (cmd.while
>                (expr.op (if signed then bopname.lts else bopname.ltu)
>                         (expr.var from_var)
>                         (expr.var to_var))
>                body_impl)
>             k_impl
>           <{ pred (nlet_eq vars v k) }>

(provided this is inside a real command) this is not indented as you want (see the shift of `cmd.seq`). I understand this is a hoare-like triple?
I will try to give a solution, but you were relying on a bug :-): basically you exploited the fact that this was indented badly (think that `<` as `+`):
```
  Definition foo :=
    + 3 +
    5. (* this should be (and is now) shifted to the right*)
```

cpitclaudel:
Indeed, it's a hoare triple.  The command being indented isn't too bad.

> I will try to give a solution, but you were relying on a bug :-)

I suspected so :) I'm happy to change the syntax, too — but I was very happy when I found that trick, since it allowed me to indent very naturally:

```
Definition … :=
   <{ … }>
   Some code here
     and the continuation lines worked perfectly!
   <{ and this magically aligned with the first part of the triple :) }>
```

So the really nice part is that the two halves of the triple line up ^^   It would be fine if the program part was indented further.

monnier:
> I suspected so :) I'm happy to change the syntax, too — but I was very happy when I found that trick, since it allowed me to indent very naturally:
>
> ```
> Definition … :=
>    <{ … }>
>    Some code here
>      and the continuation lines worked perfectly!
>    <{ and this magically aligned with the first part of the triple :) }>
> ```

Actually, if you take the analogy with `+` and treat both `<` and `>` as
having the same precedence, then the above would "naturally" indent to:

```
Definition … :=
   <{ … }>
    Some code here
      and the continuation lines worked perfectly!
   <{ and this magically aligned with the first part of the triple :) }>
```


        Stefan

Matafou:
@monnier you are right but only if `<` and `>` are declared left associative. I think I will do this. It behaves the way  @cpitclaudel  wants. It is still a hack that said. I wonder if there could be a better way. Maybe a special "three parts" expression...

Matafou:
@cpitclaudel the `!fnspec ... / ... , ... ` gets a shift to the right after the `,`. Is it ok?

cpitclaudel:
> @cpitclaudel the `!fnspec ... / ... , ... ` gets a shift to the right after the `,`. Is it ok?

Yes, that's even better than the current indentation actually :)

monnier:
> @monnier you are right but only if `<` and `>` are declared left
> associative.

Or just "associative" (i.e. have the same left and right precedence), as
is used for `;`, `,`, etc..

> I think I will do this.  It behaves the way  @cpitclaudel wants.
> It is still a hack that said.

Yup.

> I wonder if there could be a better way.  Maybe a special "three
> parts" expression...

I think doing it right with Clément's current syntax requires either
changing the lexer (so it returns different tokens for the "<{" and the
"}>" of the preconditions than those of the post conditions) or extra
ad-hoc rules in the indentation rules: the grammar itself can't be made
to understand the structure because it is basically

    assert ::= "<{" e "}>"
    e ::= ... | assert e assert

but operator precedence grammars have as core constraint that you need
a terminal between every pair of non-terminals.  You can turn the grammar
into:

    e ::= ... | "<{" e "}>" e "<{" e "}>"

but this leads to unsolvable constraints, the closest solution to those
constraints is to make those tokens "associative" (i.e. same precedence
on both sides and same precedence for both tokens), so you lose the
original structure.

My guess is that if you want to indent the middle `e` deeper, the better
solution is to add an ad-hoc indenting rule (e.g. indent deeper if
you're right after a "}>" and the next token has higher left precedence).


        Stefan


PS: This is also why you can't really have `.` as terminator but only
as separator:

    terms ::= term
    terms ::= term "." terms

is fine but

    term1 ::= term "."
    terms ::= term1
    terms ::= term1 terms

isn't since there's no terminal between `term1` and `terms`.

cpitclaudel:
Thanks both!  I'm fine with changing the syntax if that helps, the current one was co-evolved with the previous parser specifically so that it would indent nicely :)

Matafou:
> > @monnier you are right but only if `<` and `>` are declared left associative.

> Or just "associative" (i.e. have the same left and right precedence), as is used for `;`, `,`, etc..

That is not what I see. if I have `(assoc "<" ">")` I have this indentation:
```
  Definition foo :=
    < fooo
        bar >
      rab
      < foo
          bar >.
```
Maybe there is something wrong with my grammar? It says simply `(exp "<" exp) (exp ">" exp)`.
Only with "left" do I have the two "<" at the same column.

Matafou:
@erikmd can you tell me how to have the new indentation tests (currently in `ci/test-indentation`) should be made part of the CI?

Not a blocker anyways.

monnier:
>> > @monnier you are right but only if `<` and `>` are declared left associative.
>> Or just "associative" (i.e. have the same left and right precedence), as
>> is used for `;`, `,`, etc..
> That is not what I see. if I have `(assoc "<" ">")` I have this indentation:
> ```
>   Definition foo :=
>     < fooo
>         bar >
>       rab
>       < foo
>           bar >.
> ```

Hmm... you might be right that the rule that does what I expected checks
that the token is equal rather than just checking the precedence.

erikmd:
> @erikmd can you tell me how to have the new indentation tests (currently in `ci/test-indentation`) should be made part of the CI?
> 
> Not a blocker anyways.

Hi @Matafou, thanks for all this work on the indentation engine!
I've just seen your mention about CI, basically the place where you committed the indentation test-suite is very fine (in `ci/`), and we'll just need to add some "GitHub Actions job" + some shell commands in file `.github/workflows/test.yml` to launch the tests themselves automatically (and fail if an error has been detected).

Indeed, I guess it'd be very nice if this automation can be added before merging the PR (if you want we'll be able to discuss this briefly in the upcoming PG telco).

erikmd:
Hi @Matafou, I've just run the `tests.sh` and I got this:
```
Testing indent-tac-boxed.v... TEST FAILURE ***
  *** details can be seen with:
  ***  diff --side-by-side --suppress-common-lines indent-tac-boxed.v indented_indent-tac-boxed.v
  *** or graphically: 
  ***  meld indent-tac-boxed.v indented_indent-tac-boxed.v
Testing indent-inside-command-boxed.v... TEST FAILURE ***
  *** details can be seen with:
  ***  diff --side-by-side --suppress-common-lines indent-inside-command-boxed.v indented_indent-inside-command-boxed.v
  *** or graphically: 
  ***  meld indent-inside-command-boxed.v indented_indent-inside-command-boxed.v
Testing indent_monadic.v... TEST FAILURE ***
  *** details can be seen with:
  ***  diff --side-by-side --suppress-common-lines indent_monadic.v indented_indent_monadic.v
  *** or graphically: 
  ***  meld indent_monadic.v indented_indent_monadic.v
Testing indent-inside-command.v... TEST FAILURE ***
  *** details can be seen with:
  ***  diff --side-by-side --suppress-common-lines indent-inside-command.v indented_indent-inside-command.v
  *** or graphically: 
  ***  meld indent-inside-command.v indented_indent-inside-command.v
Testing indent-tac.v... TEST FAILURE ***
  *** details can be seen with:
  ***  diff --side-by-side --suppress-common-lines indent-tac.v indented_indent-tac.v
  *** or graphically: 
  ***  meld indent-tac.v indented_indent-tac.v
Testing indent-commands-boxed.v... TEST FAILURE ***
  *** details can be seen with:
  ***  diff --side-by-side --suppress-common-lines indent-commands-boxed.v indented_indent-commands-boxed.v
  *** or graphically: 
  ***  meld indent-commands-boxed.v indented_indent-commands-boxed.v
Testing indent-commands.v... TEST FAILURE ***
  *** details can be seen with:
  ***  diff --side-by-side --suppress-common-lines indent-commands.v indented_indent-commands.v
  *** or graphically: 
  ***  meld indent-commands.v indented_indent-commands.v
Testing indent_equations.v... TEST FAILURE ***
  *** details can be seen with:
  ***  diff --side-by-side --suppress-common-lines indent_equations.v indented_indent_equations.v
  *** or graphically: 
  ***  meld indent_equations.v indented_indent_equations.v
```
Attaching a zip of the folder with `*.log` files.

[test-indent.zip](https://github.com/ProofGeneral/PG/files/6850776/test-indent.zip)

Matafou:
@erikmd I can't reproduce, not easy for me to install emacs-26.
@hendriktews can you check that with 27 it works and 26 it does not please?

Matafou:
I will merge today if noone objects.

Matafou:
Let us wait for bug reports now :-).

PR #601, opened 2021-09-18
fix: parsing of `coqtop -v` for 8.14+rc1

### LINKED ISSUE #600: coq-unclassifiable-version error for +rcN version suffix
According to Coq [CEP 52](https://github.com/coq/ceps/blob/master/text/052-platform-release-cycle.md), Coq no longer has beta versions, but instead *release candidates*, with version names according to the format `X+rcN`, for example, `8.14+rc1`. We just added a package with version `8.14+rc1` to the `core-dev` Coq opam repository.

However, this new version scheme seemingly breaks ProofGeneral. Here is the error I get whenever I start up emacs with the latest Melpa package:
```
File mode specification error: (coq-unclassifiable-version . 8.14+rc1
compiled)
```

I'm guessing there needs to be special support for the `+rcN` format, like there was for `+betaN`.

erikmd:
Close #600 

Note: the function `(coq--version<)` already supported `_+rcN` suffixes.

So the compatibility issue was just coming from the fact `coqtop -v`
changed its output format; namely:

The Coq Proof Assistant, version 8.13.2 (September 2021)
compiled on Sep 12 2021 11:24:20 with OCaml 4.07.1

The Coq Proof Assistant, version 8.14+rc1
compiled with OCaml 4.12.0

### DIFF
--- a/coq/coq-system.el
+++ b/coq/coq-system.el
@@ -123,7 +123,6 @@ This function supports calling coqtop via tramp."
           (if (or (not expectedretv) (equal retv expectedretv))
               (buffer-string)))
       (error nil))))
-        
 
 (defun coq-autodetect-version (&optional interactive-p)
   "Detect and record the version of Coq currently in use.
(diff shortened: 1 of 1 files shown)


erikmd:
The CI (with all other Coq versions ≤ 8.13) is still fine, so I'll now merge this hotfix.

PR #604, opened 2021-10-03
fix(generic/pg-user.el): Auto-indent within `C-c C-n`

### LINKED ISSUE #603: SO question: How can I get `C-c C-n` to format the current line in proof-general coq-mode-map
Reported by [`@azani`](https://stackoverflow.com/users/1243445/azani) on [StackOverflow](https://stackoverflow.com/q/69419117/9164010):

> I'm using proof-general to write Coq proofs.
> 
> When I use `C-c C-n` in a proof, my cursor is moved to the next line, but the current line is not formatted. So for instance, if I type:
> 
> ```
> intros n. <C-c C-n>
> ```
> 
> my cursor moves to the next line, but `intros n.` remains unindented. So I have to go back up a line, go to the end of the line and hit `<RET>` to auto-indent the coq statement. At which point, `proof-general` considers that statement changed and I have to re-run it.
> 
> Ideally, I would line `C-c C-n` to auto-indent the line it is running and to not go to the next line.
> 
> How can I make that happen?

erikmd:
* (proof-assert-next-command-interactive): Call (proof-indent-line) by
  default or (smie-indent-line) if the current proof-assistant is coq.

Close #603

### DIFF
--- a/generic/pg-user.el
+++ b/generic/pg-user.el
@@ -192,6 +192,9 @@ If inside a comment, just process until the start of the comment."
    (save-excursion
      (goto-char (proof-queue-or-locked-end))
      (skip-chars-forward " \t\n")
+     (if (eq proof-assistant-symbol 'coq)
+         (smie-indent-line)
+       (proof-indent-line))
      (proof-assert-until-point))
    (proof-maybe-follow-locked-end)))
 


cpitclaudel:
This sounds like the perfect candidate for advice, instead of adding it directly to the library; but even if we want it in PG, I'd like to first understand the use case, which sounds like an AB problem to me.

(Oh, and this should be made optional; it's an important feature to me that stepping interactively through an improperly indented proof does *not* reindent it — there a separate button `TAB` for this)

erikmd:
Hi @cpitclaudel, thanks for your feedback. So:

> this should be made optional; it's an important feature to me that stepping interactively through an improperly indented proof does not reindent it — there a separate button TAB for this

Fully agree.

> This sounds like the perfect candidate for advice,

I mostly agree, but I believe that `advice-add` wouldn't work perfectly if we issue `C-c C-n` while the point is not exactly at the end of the locked region (hence the specific place where I added the `*-proof-indent` in this PR).

> instead of adding it directly to the library; but even if we want it in PG, I'd like to first understand the use case, which sounds like an AB problem to me.

OK, so I guess the underlying "use case" / "bug" is that reindentation has a side-effect on the handling of the locked region:
assuming we processed a given proof (e.g. with `C-c C-n`), doing `RET` on this line backtracks.
I really believe this behavior is distracting and should be fixed.
But admittedly, this is not always related to `C-c C-n` as it can also occur if we just did `C-c C-RET`.

So in the end, I guess this small PR should not be merged, in favor of a more precise fix, related to backtracking.
(But I won't have the time to implement this right now.)

WDYT?

cpitclaudel:
I agree: it's the general problem that adding space (and in most cases removing it) should not cause backtracking.

monnier:
> * (proof-assert-next-command-interactive): Call (proof-indent-line) by
>   default or (smie-indent-line) if the current proof-assistant is coq.

I think `indent-according-to-mode` is the better choice.

Also, I agree with Clement that by default `C-c C-n` should not modify
the buffer.

Regarding the undesired retraction when you reindent the line after `C-c
C-n`, maybe we should always refrain from retracting anything in response
to reindentation, something like the patch below.


        Stefan


diff --git a/generic/proof-script.el b/generic/proof-script.el
index 44a0cb34..07010d86 100644
--- a/generic/proof-script.el
+++ b/generic/proof-script.el
@@ -3,7 +3,7 @@
 ;; This file is part of Proof General.
 
 ;; Portions © Copyright 1994-2012  David Aspinall and University of Edinburgh
-;; Portions © Copyright 2003-2018  Free Software Foundation, Inc.
+;; Portions © Copyright 2003-2021  Free Software Foundation, Inc.
 ;; Portions © Copyright 2001-2017  Pierre Courtieu
 ;; Portions © Copyright 2010, 2016  Erik Martin-Dorel
 ;; Portions © Copyright 2011-2013, 2016-2017  Hendrik Tews
@@ -2024,10 +2024,13 @@ This function expects the buffer to be activated for advancing."
     (proof-script-delete-secondary-spans startpos lastpos)
     (proof-extend-queue lastpos vanillas)))
 
+(defvar proof-inhibit-retract-on-change nil)
+
 (defun proof-retract-before-change (beg end)
   "For `before-change-functions'.  Retract to BEG unless BEG and END in comment.
 No effect if prover is busy."
   (when (and (> (proof-queue-or-locked-end) beg)
+	     (not proof-inhibit-retract-on-change)
 	     (not (and (proof-inside-comment beg)
 		       (proof-inside-comment end))))
     (when proof-shell-busy
@@ -2651,6 +2654,11 @@ finish setup which depends on specific proof assistant configuration."
       (proof-shell-make-associated-buffers)
       (proof-layout-windows))
 
+  (add-function :around (local 'indent-line-function)
+                #'(lambda (orig-fun &rest args)
+                    (let ((proof-inhibit-retract-on-change t))
+                      (apply orig-fun args))))
+
   ;; Make sure the user has been welcomed!
   (proof-splash-message))

Matafou:
Indeed indentation could be an exception to the locked region like editing comments is.
But in this precise case one can maybe decide to indent *before* sending to the prover and locking. In case of error it would maybe be odd.

Matafou:
> * @Matafou I don't know what you mean by "in case of error it would maybe be odd."

If we indent the line before scripting/locking it, then if the line causes an error we have indented a line without locking it. That may seem a bit odd to the user.

PR #606, opened 2021-10-07
* generic/proof-script.el: Don't retract on indent (#604)

monnier:
(proof--inhibit-retract-on-change): New var.
(proof-retract-before-change): Obey it.
(proof-config-done): Set it while indenting.

### DIFF
--- a/coq/coq.el
+++ b/coq/coq.el
@@ -2732,10 +2732,8 @@ insertion point for the \"using\" annotation. ")
 
 (defun coq-insert-proof-using (_proof-pos _previous-content insert-point string-suggested)
   (goto-char insert-point)
-  (let ((spl proof-locked-span))
-    (span-read-write spl) ; temporarily make the locked span writable
-    (insert (concat " using " string-suggested))
-    (proof-span-read-only spl)))
+  (let ((proof--inhibit-retract-on-change t))
+    (insert (concat " using " string-suggested))))
 
 (defun coq-insert-suggested-dependency ()
   (interactive)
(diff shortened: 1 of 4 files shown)


Matafou:
LGTM. Will merge soon if no objection.

cpitclaudel:
FWIW, this or something around that time broke PG pretty badly for me.

Right now when I go back and forth over a proof Proof General frequently gets confused and displays a processed region with incorrect boundaries.  In fact, it looks as if there's some sort of overlay cache that gets reused when moving forward.  And it's not just a visual issue: it's so bad that PG sometimes ends up sending stale stuff to Coq.  Any ideas?

cpitclaudel:
Yep, it's a caching issue.

I have this input:

image

I press `C-c C-RET`, and I get

image

And this is because of stale positions in PG's cache:

image

cpitclaudel:
And indeed `(setq proof-use-parser-cache nil)` fixes the issue.  We should either turn the cache off, or investigate further.

Matafou:
Strange indeed. I have never seen this problem. I use PG on a daily basis, but most of the time on small files. I od get back and forth a lot though.
OK to unplug the cache while we investigate.

Matafou:
Do you really think it has something to do with this PR?

Matafou:
It may be that the whole "edit the locked region" is incompatible with the cache?

cpitclaudel:
> It may be that the whole "edit the locked region" is incompatible with the cache?

Pretty much, yes: we know that's possible (since it works when we do it in comments), but there has to be a special thing do to to tell PG about the changes.  Yet looking at the code I don't see anything special there...

I can't swear that it's this commit, but I upgraded to this commit recently and things have been pretty broken.

Matafou:
Do you often use the ability to indent in the lock region?

cpitclaudel:
es, a lot: it happens automatically with aggressive indent mode, and the new indentation engine has a lot of cases where it doesn't do what I expect, so it reindents.

Matafou:
This makes me suspect that this bug is much older than this PR, we probably almost never triggered it before.

PR #607, opened 2021-10-11
Fix #597; ProofGeneral cannot step over `Fail` correctly

### LINKED ISSUE #597: ProofGeneral cannot step over `Fail` correctly when `Set Ltac Backtrace` is enabled
On my (Spacemacs) installation, if `Set Ltac Backtrace` is enabled, `Fail` still works correctly in Coq (8.13.2), but ProofGeneral does not seem to understand that. The same example works in VsCoq.

```coq
Goal False.
Proof.
  Fail (now auto). (* Stepping to here succeeds. *)
  auto. (* Stepping to here succeeds. *)
Abort.

#[local] Set Ltac Backtrace.
Goal False.
  Fail (now auto). (* Stepping to here fails. *)
  auto. (* Stepping to here succeeds. *)
Abort.
```

I was alerted to this interaction by @jhaag (but his symptoms are different, so questions remain).

Matafou:
The bug happened when `Set Ltac Backtrace` is enabled, which has
become the default since a few versions. The fix is to prefix the
coq-error-message variable with a negative regexp for "message:\n".

This PR should be accepted if it does not slow performances too much.

- [x] check performance
- [x] Add tests

### DIFF
--- a/ci/coq-tests.el
+++ b/ci/coq-tests.el
@@ -303,6 +303,44 @@ For example, COMMENT could be (*test-definition*)"
    'show-proof-stepwise 'diffs-on))
  
 
+(ert-deftest 090_coq-test-regression-Fail()
+  "Test for Fail"
+  (coq-fixture-on-file
+   (coq-test-full-path "test_stepwise.v")
+   (lambda ()
+     (coq-test-goto-before "(*FailNoTrace*)")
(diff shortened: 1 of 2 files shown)


Matafou:
Performance seem OK.
Tests added (special case with coq 8.9).
Ready to merge.

PR #609, opened 2021-10-12
Fix #608. PG not auto adapting window width.

### LINKED ISSUE #608: Coq Printing Width Does not Set to window size
Hi,

I just updated my proof general and it turns out the `Coq Goals` buffer no longer auto adjust the output width to the window size but default to 80, even `coq-adapt-printing-width` is set to the current window size.

I have to do the following to fix this issue:
1. `C-c C-a C-w` and set it to 80
2. `C-c C-a C-w` and set it to the current window size

My environment:
Proof General 4.5-git
coq 8.4pl6 (October 2020)
Spacemacs v.0.300.0
company-coq 20210708.2357 
Emacs 27.1(9.0), installed from homebrew(`brew tap railwaycat/emacsmacport && brew install emacs-mac`)
macOS Big Sur 11.6

Any help would be appreciated!

Matafou:
The option setting was flawed. A recent code cleaning by @monnier revealed it.
Now the hooks are set once and forall correctly and obey the setting.

- [ ] add tests

### DIFF
--- a/ci/coq-tests.el
+++ b/ci/coq-tests.el
@@ -133,7 +133,9 @@ then evaluate the BODY function and finally tear-down (exit Coq)."
 ;;; For info on macros: https://mullikine.github.io/posts/macro-tutorial
 ;;; (pp (macroexpand '(macro args)))
   (save-excursion
-    (let* ((openfile (or file
+    (let* (;; avoids bad width detection in batch mode 
+           (coq-auto-adapt-printing-width nil)
+           (openfile (or file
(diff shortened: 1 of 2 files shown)


Matafou:
The changes in the "auto adapt width" feature had an unexpected effect. In batch mode now by default the width is considered very small. Hence if I don't set explicitly `coq-auto-adapt-printing-width` to `nil` the tests fail.

Matafou:
I don't really know how to test this feature in batch mode. How to set the width of a window that don't really exist?

PR #626, opened 2021-11-17
test various Coq and Emacs versions using docker-coq-action

hendriktews:


hendriktews:
@erikmd Here is another try, this time based on docker-coq-action.

erikmd:
Hi @hendriktews, thanks,
But I didn't understand at first sight why you refactored the whole GHA configuration so deeply 😅
IINM, the new images should essentially allow us to be more flexible about the (coq, emacs) versions…
while keeping the scripts themselves as is (or so)
Anyway let me know if I can be of some help with your experiments this week!

erikmd:
BTW @hendriktews you might be interested in the syntax of `include:` or `exclude:`, maybe
(doc'ed in https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix)

e.g.:
```yaml
strategy:
  matrix:
    coq: ['8.14', '8.13', '8.12', '8.11', '8.10', '8.9', '8.8', '8.7', '8.6']
    emacs: ['27.2', '25.1']
    include:
      - coq: '8.14'
        emacs: '26.1'
```

hendriktews:
I'll have a look at include and exclude, thanks for the hint. I experimented with anchors, but they are not supported.

> Hi @hendriktews, thanks, But I didn't understand at first sight why you refactored the whole GHA configuration so deeply

I am confused, I don't think I refactored. I just temporarily deleted many jobs here to make the experiments faster. Or what are you referring to?

erikmd:
Hi @hendriktews,

> I experimented with anchors, but they are not supported.

Indeed, I confirm YAML anchors are not supported by GitHub Actions for the time being.

If ever one wants to "factor" things within a single workflow file, we need to either (1) call an external script file or (2) use [global environment variables (at workflow level)](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idenv)

> I don't think I refactored. I just temporarily deleted many jobs here to make the experiments faster.

Indeed sorry, I had misinterpreted this at first 😅

hendriktews:
OK, I am working now on a matrix which coq/emacs versions to test. Let's see how it will evolve.

hendriktews:
discussed this PR in the maintainer telco, this is ready to be merged now

PR #628, opened 2021-11-20
proof-general-pkg.el: Let it be auto-generated

monnier:


erikmd (approved):
LGTM, so merging this PR just now.

And tomorrow morning, we'll see if the automatic MELPA deployment is still fine at
https://melpa.org/#/proof-general
and still mentions `emacs 25.1` as a dependency.

erikmd:
Forget what I said:

it seems it is OK because https://melpa.org/#/proof-general mentions https://github.com/ProofGeneral/PG/tree/2376485828bbd6f151897fdac77dab84f360100e

I was puzzled just because the date was still `20211109.1442` (because the last merge did not touch the business logic, maybe)

monnier:
> I believe your "cleanup improvement" didn't work:
>
> I merge the PR 14h ago, but proof-general is no more built apparently:
> https://melpa.org/#/proof-general
>
> Do you believe we need to revert your PR #628 and see if this fixes the issue?

I'm not sufficiently familiar with MELPA to know exactly what's the problem.
Maybe the recipe needs to be updated to include `proof-general.el` in
the list of `:files`?

[ On the positive side,
  http://elpa.nongnu.org/nongnu-devel/proof-general.html did build the
  new version.  ]

erikmd:
Actually @monnier as I had told in my very last message,
the build is OK because the commit SHA1 is the proper one.

And I just believe the "date" that is shown on MELPA is not the "CommitDate / AuthorDate" of the last commit, but the last "CommitDate" of the last changed file… (and as the last merge did not impacted an existing file but removed one, maybe that's the explanation).

To sum up: we shouldn't worry about MELPA now; we'll just see after the upcoming merge if my current guess was OK.

PR #631, opened 2021-11-29
CI workflow: add Emacs 27.2

hendriktews:


hendriktews:
Abandoning because ` ci/coq-tests.el` does not run with emacs 27.2 here.

PR #632, opened 2021-11-30
Improve CI workflow

hendriktews:


erikmd:
Hi @hendriktews!

Happy New Year 2022 ✨✨✨

Just FYI: Docker-Coq now gathers `coqorg/coq:8.15-ocaml-*` images.
So maybe you may want to extend this PR accordingly.

hendriktews:
I extended this to 8.15 and 27.2. Note that this PR is just a conservative extension of the old CI workflow. It doesn't use containers and it tests all Coq versions just with one emacs version.
See #626 for a PR that adds containers and tests different Coq/Emacs version pairs.
This could be merged without much discussion, if there were not #633.

hendriktews:
With 8.15, there is also a compile test failing, will look into this.

hendriktews:
I believe the compile test is failing because of a coq problem, see coq/coq#15773.

hendriktews:
The vok issue in 8.15.0 does not leave any other choice but to expect failures for tests that check vok files.

erikmd:
OK, thanks @hendriktews !

PR #639, opened 2022-02-11
Add an option to stop proof-shell-kill-function from killing buffers.

Chobbes:
I can't be the only one that is bothered by this! Every time you switch between files proof general kills and restarts the goal and response buffers. This is particularly painful if you have `proof-multiple-frames-enable` set --- I want to let xmonad handle the windows instead of emacs, but then every time I edit a new file the frames are deleted and new ones are created, messing up the layout. This PR adds a custom variable to disable this behaviour.

Happy to make changes if necessary!

### DIFF
--- a/generic/proof-shell.el
+++ b/generic/proof-shell.el
@@ -494,6 +494,22 @@ process command."
 (defvar proof-shell-kill-function-hooks nil
   "Functions run from `proof-shell-kill-function'.")
 
+(defun proof-shell-kill-function-kill-associated-buffers ()
+  "Kill the goal, response, and trace buffers, as well as frames if necessary."
+  ;; Remove auxiliary windows, trying to stop proliferation of
+  ;; frames (NB: loses if user has switched buffer in special frame)
+  (if (and proof-multiple-frames-enable
+	   proof-shell-fiddle-frames)
+      (proof-delete-all-associated-windows))
+
+  (let ((proof-shell-buffer nil)) ;; fool kill buffer hooks
+    (dolist (buf '(proof-goals-buffer proof-response-buffer
+				      proof-trace-buffer))
+      (when (buffer-live-p (symbol-value buf))
+        (delete-windows-on (symbol-value buf))
+        (kill-buffer (symbol-value buf))
+        (set buf nil)))))
+
 (defun proof-shell-kill-function ()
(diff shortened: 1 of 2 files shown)


Chobbes:
Hey, just wanted to poke this to see if there's anything blocking this being merged. Happy to make changes as necessary.

hendriktews:
Thanks a lot for your contribution. I will discuss this PR with the other PG maintainers. Please be patient meanwhile.

Matafou:
This PR seems fine to me. Althoug I did not test it. In particular we should check that the following still work correctly:

- [ ] coqtop process is always re-launched when one changing the script buffer
- [ ] coq arguments are always re-computed when one changing the script buffer
- [ ] maybe check that these buffers are blanked?

Chobbes:
From my own testing:

- [x] coqtop process is always re-launched when changing the script buffer
- [x] buffers (response / goal / etc) are blanked

Not sure what you mean by coq arguments being recomputed... I can confirm that coqtop gets a new process and new arguments for `-topfile` and friends when switching files. I have no problems switching between coq projects.

yiyunliu:
For the bullet point: 
> coq arguments are always re-computed when one changing the script buffer
I was able to switch successfully between two projects with different -Q options (ctrl + enter to a file from the second project when I'm in an interactive session for a file from the first project). Is that sufficient to show this part is handled correctly? I believe the switch would have failed if coqtop were passed the same arguments.

Matafou:
That is indeed what I had in mind. Thanks for the confirmation.

Matafou:
I tested and everything seems fine afaiks. Thanks a lot for you contribution! Actually I wonder why we did not do this before.

In single frame mode this has no real use case for now: current window layout will not  be kept next time coq is started (`proof-layout-windows` is called). We should maybe think about having a way to preserve window layout between coq sessions. 
Anyway I merge this. Thanks again!

Chobbes:
Yeah! This always bothered me. Really annoying how the layout gets destroyed constantly. At least now you can preserve it as long as you don’t close emacs… there are ways to preserve window configurations in emacs, even in registers, so I’m sure it’s possible to remember a layout.

But yeah, killing the buffers never made sense to me, I almost wonder if we should default to not killing them. Maybe it made more sense in the past, or with other proof assistants.

Calvin Beck

Issue #621, opened 2021-11-09
Feature requests for the new indentation engine


cpitclaudel:
Hi @Matafou 

Thanks for your work on the new indentation engine!  Here are a few things that could be improved.

The following samples are indented how I would expect them; PG's new indentation changes them:

```
Instance spec_of_upstr : spec_of "upstr" :=
  fnspec! "upstr" s_ptr wlen / (s : list byte) R,
    { requires tr mem :=
        wlen = word.of_Z (Z.of_nat (length s));
      ensures tr' mem' :=
        tr' = tr }.
```

```
Goal forall x,
    x =
    x.
```

```
Goal (True /\
      True /\
      True).
```

```
Lemma t :
  (forall x,
     x = x).
Qed.
```

```
match x with
| ABC =>
  DEF
end
```

monnier:
> ```
> Goal forall x,
>     x =
>     x.
> ```

While I can agree with (or understand) the other choices, FWIW, the
above looks wrong to me.


> ```
> Goal (True /\
>       True /\
>       True).
> ```

And here, I'll note that the GNU coding convention recommends breaking
the line *before* infix operators rather than after them.  I found it
odd, to be honest, but it's growing on me (tho still definitely not for
the `,` and `;`).

cpitclaudel:
> While I can agree with (or understand) the other choices, FWIW, the
above looks wrong to me.

That's how PG used to do it; not sure if it's the "right" way, but it lines things up nicely (often you have roughly the same thing before and after the equal, modulo some tweak)

cpitclaudel:
> And here, I'll note that the GNU coding convention recommends breaking
> the line *before* infix operators rather than after them.

yes, I dislike this convention for the same reason: things don't line up:

```
uint32_t x = ((x0 << 24)
              | (x1 << 16)              
              | (x2 <<  8) 
              | (x3 <<  0));
```

vs.

```
uint32_t x = ((x0 << 24) |
              (x1 << 16) |
              (x2 <<  8) |
              (x3 <<  0));
```

Dafny does it better:

```
uint32_t x = (| (x0 << 24)
              | (x1 << 16)              
              | (x2 <<  8) 
              | (x3 <<  0));
```

(You're allowed to prefix one copy of the infix operator in a chain, which works nicely for `&&` in particular)

Matafou:
Hi! Nice to have feedback!

> Hi @Matafou
> 
[...]

PR #647, opened 2022-03-25
Fix #646. Indentation of function arguments.

### LINKED ISSUE #646: indentation of function application arguments
[from here](https://coq.zulipchat.com/#narrow/stream/304019-Proof-General-users/topic/Better.20emacs.20indentation/near/275291823).

It is indeed quite annoying that in function applications arguments on there own line line up with the first argument instead of the function name:

```coq
Definition foo :=
  my_function_name a b
                   c d.
```
is really ugly and space-expensive compared to: 
```coq
Definition foo :=
  my_function_name a b
    c d.
```
I think the previous indentation implementation used to have the latter.

Matafou:
Fix #646.  Now indentation of function arguments are indented wrt to the function instead of aligned with the first argument. See details below.

## Questions

- Is it too late for the including this in the soon to come V.4.5 tag? The indentation CI test are a good clue that this does not break indentation too much but we may wait for users to test this a bit.
- should we make the old behaviour still possible with a switch? I don't thinks `coq-indent-box-style` would be the right switch. I am not found of addig one more switch.

## Details

This patch indents like this:
```
Definition foo :=
  myfunction (args1) arg2
    arg3.
```
instead of 
```
Definition foo :=
  myfunction (args1) arg2
             arg3.
```

Besides when a function application is placed immediately after (and on the same line as) a "Com start" token, indentation is computed from the "Com start" instead. Because in practice the command start is the real function call. eg.:

```
Require Import
  Foo.
```
instead of 
```
Require Import
        Foo.
```
A few tests have been added to check this new cases.
A few tests have been modified to reflect the change.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -204,6 +204,19 @@ also https://wiki.ubuntu.com/Releases).
     Variable `coq-indent-box-style' only affects indentation after
     quantifiers (used to affect script braces "{").
 
+    Variable `coq-indent-align-with-first-arg` governs between these
+    two indentation behaviours:
+
+    (setq 'coq-indent-align-with-first-arg t) gives
+
+        somefunciton x y
+                     z t u
+                     v
+    (setq 'coq-indent-align-with-first-arg nil) give (default):
+
+        somefunciton x y
+          z t u
+          v
 
 * Changes of Proof General 4.4 from Proof General 4.3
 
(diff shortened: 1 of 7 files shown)


Matafou:
also implements one of the feature wishes of #621 for indenting records ":=" asymmetrically.

Matafou:
pushed a few more comments.

PR #648, opened 2022-03-28
Fix #599, bad detection of "Proof" in "Proof using" suggestions.

### LINKED ISSUE #599: `coq-accept-proof-using-suggestion 'always` incorrectly believes that `pose proof` is a proof directive
Hence it makes suggestions like
```diff
-    Lemma width_nonneg : 0 <= width. pose proof width_pos. blia. Qed.
+    Lemma width_nonneg : 0 <= width. pose proof using word_ok width_pos. blia. Qed.
```
which is nonsense

Matafou:
Also added a test. Not sure which versions of coq support this. Let us see with CI.

### DIFF
--- a/ci/coq-tests.el
+++ b/ci/coq-tests.el
@@ -343,6 +343,27 @@ In nested Ltac calls to \"now (tactic)\" and \"easy\", last call failed.
 Tactic failure: Cannot solve this goal."))))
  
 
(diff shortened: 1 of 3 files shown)


Matafou:
OK. All tested versions of coq succeed. Will merge soon.
The test only checks that the right "Proof" keyword is highlighted.

PR #649, opened 2022-03-28
Fix #618 - Support for command "Derive SuchThat As".

### LINKED ISSUE #618: Feature request: indentation of Derive … SuchThat
Our indentation doesn't support `Derive`, so instead of this:

```
Derive X SuchThat
  P
As U.
Proof.
```

we produce this:

```
  Derive X SuchThat
         P
         As U.
  Proof.
```

Here's another test:

```
  Derive X
  SuchThat P
  As U.
  Proof.
```

which we indent as

```
  Derive X
         SuchThat P
         As U.
  Proof.
```

(we should also add syntax highlighting for Derive)

Matafou:
Indentation + syntax highlighting.


cpitclaudel:
Thanks!!

PR #650, opened 2022-03-28
test: Replace `omega` with `lia` in CI & Add Coq versions

Matafou:
Testing if replacing lia with omega in test suite is OK for all old versions. then that would fix the coq-8.15 test failure (deprecated omega was removed in coq-8.15).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -93,6 +93,8 @@ jobs:
           - '8.11'
           - '8.12'
           - '8.13'
+          - '8.14'
+          - '8.15'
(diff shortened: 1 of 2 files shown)


Matafou:
@erikmd  @hendriktews this PR fixes some errors in tests, but not everything. I am not sure what the remaining errors are.

erikmd:
Hi! sorry for replying late.
The good news is that the CI is not fully green: `coqorg/coq:8.15` now contains Coq 8.15.1 which fixes the issue Hendrik had reported upstream.
So, merging now.

PR #654, opened 2022-05-23
add customizable variables to configure which part of a goal the goal buffer will focus on

cipher1024:
By setting `coq-prefer-top-of-conclusion` to `t`, we get Proof General to focus on the conclusion of the first goal after displaying the list of goals of a proof. The conclusion is defined as whatever is preceded by `=======`. In some cases, it can be useful to mark a different part of the conclusion as the main point of interest. This PR allows the customization of that marker as well as the ability to choose where to place that marker in the window when focusing on it.

### DIFF
--- a/coq/coq.el
+++ b/coq/coq.el
@@ -117,7 +117,7 @@ Namely, goals that do not fit in the goals window."
 
 (defcustom coq-goal-conclusion-marker "========"
   "Marks the part of interest in each goal to focus the goal buffer on"
-  :type regexp
+  :type 'regexp
   :group 'coq
   )
 
--- a/coq/coq.el
+++ b/coq/coq.el
@@ -115,6 +115,18 @@ Namely, goals that do not fit in the goals window."
   :type 'boolean
   :group 'coq)
 
+(defcustom coq-goal-conclusion-marker "========"
+  "Marks the part of interest in each goal to focus the goal buffer on"
+  :type regexp
+  :group 'coq
+  )
+
+(defcustom coq-center-goal-after-focus 'top
+  "Where to display the goal marker after rendering a proof goal"
+  :type '(radio
+    (const :tag "move conclusion marker to the top of the window" top)
+    (const :tag "move conclusion marker to the middle of the window" middle)
+    (const :tag "move conclusion marker to the bottom of the window" bottom)))
 
 (defconst coq-shell-init-cmd
   (append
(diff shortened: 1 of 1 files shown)


Matafou:
Thanks. This is a good idea. I will have a look at your patch asap. Can you explain how the user does specify the part of the goal to follow? Is it by selecting some peace of it?

erikmd:
@Matafou @cipher1024 what is the status of this PR?

Matafou:
Looks good and useful to me. Thanks again!
@cipher1024 can you add a line to the CHANGES file, rebase into one commit please?
We need to add a documentation on these variables (`coq-prefer-top-of-conclusion` itself isn't).

coq/coq.el
```diff
@@ -115,6 +115,18 @@ Namely, goals that do not fit in the goals window."
...
+  :type 'regexp
+  :group 'coq
+  )
+
+(defcustom coq-center-goal-after-focus 'top
+  "Where to display the goal marker after rendering a proof goal"
+  :type '(radio
+    (const :tag "move conclusion marker to the top of the window" top)
+    (const :tag "move conclusion marker to the middle of the window" middle)
+    (const :tag "move conclusion marker to the bottom of the window" bottom)))
```
Matafou, line 130: For better user configuration should we merge this preference with `coq-prefer-top-of-conclusion`?
Currently if this one is nil then we  show the end of the conclusion (of the first goal). Something like:

```
    (const :tag "focus on the end of the conclusion even if it hides the conclusion marker" tail)
```

`coq-prefer-top-of-conclusion`could be left for compatibilty.
@cipher1024 what do you think?

cipher1024: I like it! Let's do it

coq/coq.el
```diff
@@ -3163,7 +3175,10 @@ buffer."
           (goto-char (match-beginning 0))
           (buffer-substring p (point)))))))
 
-
+(defun coq-recenter-goal (loc)
+  (cond ((eq loc 'top) (recenter 0))
+        ((eq loc 'middle) (recenter nil))
+        ((eq loc 'bottom) (recenter -1))))
```
Matafou, line 3181: let us have something like a `'overthetop` option?

Matafou:
Let us merge this and finish to polish confiuguration in another PR? 
- replace `coq-prefer-top-of-conclusion` by the `overthetop` option: focus on the top of the hypothesis (possibly hiding the conclusion marker). This would make the configuration clearer.
- have a `tail` option too to focus on the bottom of the conclusion possibly hiding the conclusion marker)

Other ideas for other PRs:
- orthogonaly to the options introduced here I think we should always prefer hiding the other goals over hiding hypothesis. So for instance when having the conclusion marker on top, if we can scroll up a bit more to show hypothesis without hiding the conclusion, we should do it.
- I always found that the number of currently focused goals is not sufficiently highlighted (it is currently shown in the modeline but it is too discrete.
- Ideally if the user scrolls the goal buffer we should probably try to remember it and limit the "re-scroll back" after each command. I don't really know this could be achieved.

hendriktews:
From glancing over it, it seems a nice thing to have. I am not opposing.

PR #659, opened 2022-06-12
CI: fix ci/coq-tests.el for emacs 28

### LINKED ISSUE #657: ci/coq-tests.el fails completely with emacs 28
See testrun on PR #626 (https://github.com/ProofGeneral/PG/actions/runs/2450648668). All test fail with `(error "Defining as dynamic an already lexical var")`. The problem is independent of the Coq version, all combinations tested with emacs 28 fail the same way.

hendriktews:
See #657, fixes #657

### DIFF
--- a/ci/coq-tests.el
+++ b/ci/coq-tests.el
@@ -19,6 +19,10 @@
 ;;(require 'ert-async)
 ;;(setq ert-async-timeout 2)
 
+;; Load Coq instance of Proof General now.
+(proof-ready-for-assistant 'coq)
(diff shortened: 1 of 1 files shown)


hendriktews:
Erik, please check if this change is fine with you. It makes running these tests interactively somewhat more cumbersome, because PG is not automatically unloaded after the test, as it was before.

hendriktews:
Discussed this change in the maintainer telco. Erik is fine with it as is.

PR #660, opened 2022-06-19
update CHANGES file

hendriktews:


hendriktews:
The 3 failing tests fail in `purcell/setup-emacs` for installing emacs. The problem seems to be transient, other tests with the same emacs version succeed.

Scutua (approved):

PR #669, opened 2022-09-06
Do not offer to save (temporary) buffers without associated files

### LINKED ISSUE #668: Interference with org-mode leads to buffer-save prompt of ephemeral buffers
When an org source block gets fontified, this happens in an ephemeral buffer with the major mode associated to the language of the block. `proof-config-done` at https://github.com/ProofGeneral/PG/blob/master/generic/proof-script.el#L2819 overzealously sets `buffer-offer-save` on such buffers. Upon exiting Emacs, this leads to the user being prompted whether they want to save the ephemeral fontification buffer, which they certainly don't want.

    #+begin_src coq
    mixin_cmra_extend n (x y1 y2 : A) :
      ✓{n} x → x ≡{n}≡ y1 ⋅ y2 →
      { z1 : A & { z2 | x ≡ z1 ⋅ z2 ∧ z1 ≡{n}≡ y1 ∧ z2 ≡{n}≡ y2 } }
    #+end_src

Why `proof-config-done` sets `buffer-offer-save` in the first place, I don't know. Temporary buffers don't get saved in Emacs; presumably a proof-general user got bitten by this once and decided to change the behaviour of Emacs by patching proof-general rather than changing their own config. I suggest the removal of that line unless there is a good reason to set `buffer-offer-save` here.

haselwarter:
This restores the default behaviour of Emacs. If a user wants to be more
careful about saving temporary buffers, they can set `buffer-offer-save` via
the appropriate hooks.

This commit effectively reverts b8cca128a40ee0d72bc627a5d320cdde12b3d87c

Closes #668

### DIFF
--- a/generic/proof-script.el
+++ b/generic/proof-script.el
@@ -2814,10 +2814,6 @@ finish setup which depends on specific proof assistant configuration."
   (proof-setup-imenu)
   (proof-imenu-enable)
 
-  ;; Save file-less script mode buffers in case of accidental exit
-  (or (buffer-file-name)
-      (setq buffer-offer-save t))
-
   ;; Turn on autosend if enabled
   (proof-autosend-enable 'nomsg)
 


Matafou:
Looks indeed bad practice. Either we let the default behaviour or we offer an optin option.
LGTM.

erikmd (approved):
LGTM as well

Matafou:
Thanks for the ping. Let us merge this right now.

PR #675, opened 2022-10-23
Miscellaneous improvements of qrhl-tool support

dominique-unruh:
In a nutshell, improvements of font-lock mode configuration, communication with qrhl-tool, indentation, and new font-lock keywords due to developments of the prover.

See the individual commit messages for details.


erikmd:
Thanks a lot, @dominique-unruh 👍

Cc @ProofGeneral/core: we should quickly proofread this PR and merge it… (unfortunately, I won't be able to do it soon enough myself)

qrhl/qrhl.el
```diff
@@ -116,8 +129,9 @@
   proof-save-command-regexp "\\`a\\`"   ;AKA `regexp-unmatchable' in Emacs-27
   proof-tree-external-display nil
   proof-script-font-lock-keywords qrhl-font-lock-keywords
-  proof-goals-font-lock-keywords qrhl-font-lock-keywords
+  proof-goals-font-lock-keywords qrhl-font-lock-subsuperscript
```
Matafou, line 140: Does this mean you print goals with a smaller font? Interesting.

qrhl/qrhl.el
```diff
@@ -116,8 +129,9 @@
...
   proof-response-font-lock-keywords qrhl-font-lock-keywords
+  font-lock-extra-managed-props '(display)
```
Matafou, line 142: Do you really need this? Is this variable buffer local? Otherwise you will mess with other modes.

Matafou:
Except for the `font-lock-extra-managed-props` question, it looks good to me.

qrhl/qrhl.el
dominique-unruh: @Matafou, no, `qrhl-font-lock-subsuperscript` is for one very specific purpose. Isabelle/HOL (which is used internally in qrhl-tool) uses `⇩` and `⇧` to make the next letter a sub- or superscript. We cannot convert this into Unicode sub/superscripts because Unicode has only a very limited set of sub/superscripts. (Actually, we do use Unicode sub/superscripts in the few cases where they exist.) So I wrote this font-lock config that replaces `⇩x` visually by a subscript `x` etc. (It's actually used not only in the goals, `qrhl-font-lock-keywords` includes `qrhl-font-lock-subsuperscript`.)

qrhl/qrhl.el
dominique-unruh: I definitely need it to support the `qrhl-font-lock-subsuperscript` (see above) because that one uses the `display` property to create sub-/superscripts. If we don't set `font-lock-extra-managed-props`, everything seems fine at first, but when editing `⇩x` into `⇩yx`, the `x` stays a subscript. (Because font-lock only updates the properties in `font-lock-extra-managed-props`.

But you are right, and I did not set it buffer local which needs to be fixed. I can do this by setting it in `qrhl-mode-hook` via `(set (make-local-variable 'font-lock-extra-managed-props) '(display))` instead of in `proof-easy-config`. It works but I am not sure it is the right place for this. (I am a bit unsure about when something belongs in the hook and when in the  `proof-easy-config`.) Please advise whether it should be put somewhere else.

erikmd:
Thanks a lot @Matafou 👍👍 (I just assigned just to "formally" mark that you'll be able to decide when to merge :)

qrhl/qrhl.el
Matafou: I think replacing `setq` by `setq-local` is enough, but I don't know how to do that with `proof-easy-config`.

Matafou: `describe-variable` on `font-lock-extra-managed-props` says that this variable becomes buffer local when set. So I guess this is OK like this.

qrhl/qrhl.el
Matafou: > (It's actually used not only in the goals, qrhl-font-lock-keywords includes qrhl-font-lock-subsuperscript.

Why would the two buffers have different fonts then?

dominique-unruh: It's not a different font. It's a different font-lock configuration for the two buffers. The script buffer has `qrhl-font-lock-keywords` which highlights keywords *and* does the sub/superscript stuff. The goal buffer has `qrhl-font-lock-subsuperscript` which only does the sub/superscript stuff but does not highlight keywords (because keywords are specific to the script language.)

qrhl/qrhl.el
dominique-unruh: @Matafou In my Emacs, it does not say so! In fact, I tried it out with the current code and the effect was global. Could it be version dependent? (I used `GNU Emacs 28.1 (build 2, x86_64-w64-mingw32) of 2022-04-21`)
```
font-lock-extra-managed-props is a variable defined in ‘font-lock.el’.

Its value is (help-echo)
Local in buffer *scratch*; global value is nil

Additional text properties managed by font-lock.
This is used by ‘font-lock-default-unfontify-region’ to decide
what properties to clear before refontifying a region.

  Probably introduced at or before Emacs version 22.1.

[back]
```
I would then suggest to use `setq-local` in the hook. Does that sound right?

Matafou: I think yes, but I also think it is time to ask @monnier :-)

qrhl/qrhl.el
Matafou: Ho I see. Thanks. In Coq we usually highlight the goals buffer, because  highlight terms and types, not only commands.

monnier:
> `describe-variable` on `font-lock-extra-managed-props` says that this
> variable becomes buffer local when set. So I guess this is OK like this.

That's odd.  AFAIK it is not "automatically buffer local".  Maybe you
have loaded a package which makes it "automatically buffer local" (that
would be a nasty thing to do, but packages have been known to do worse).

monnier:
> `describe-variable` on `font-lock-extra-managed-props` says that this
> variable becomes buffer local when set. So I guess this is OK like this.

Even if it sets it buffer-locally it would not be right, because it
would set it in the buffer that happens to be current when `qrhl.el` is
loaded rather than in the buffers where we install those qrhl
font-lock rules.

The best solution is probably to stop using `proof-easy-config`,
otherwise we'd have to extend `proof-easy-config` specifically for that.

[ Note that I'm biased because I personally dislike the kind of designs
  like `proof-easy-config` where there is no seamless evolution to more
  complex setups.  ]

dominique-unruh:
> Even if it sets it buffer-locally it would not be right, because it would set it in the buffer that happens to be current when `qrhl.el` is loaded rather than in the buffers where we install those qrhl font-lock rules.

I wanted to add it inside `qrhl-mode-hook`. As far as I understand, this would make it buffer-local in every buffer that activates `qrhl-mode`, not any other buffers. I'm adding the commit with the change for illustration. (I can remove again if it's not the right approach.)

monnier:
> I wanted to add it inside `qrhl-mode-hook`. As far as I understand, this
> would make it buffer-local in every buffer that activates `qrhl-mode`, not
> any other buffers. I'm adding the commit with the change for
> illustration. (I can remove again if it's not the right approach.)

That would work, yes.  But IIUC the new code also uses `display`
properties for super/subscripts in the response buffer, so we'd also
need that in the `qrhl-response-mode-hook`.

dominique-unruh:
@monnier 
> That would work, yes. But IIUC the new code also uses `display` properties for super/subscripts in the response buffer, so we'd also need that in the `qrhl-response-mode-hook`.

I thought about that. I didn't add it because the response buffer is never edited, only replaced. (`font-lock-extra-managed-props` is only for handling edits, not initial fontifying.) But if you think it's better to add them, too, I can do it. (There is definitely not harm in it.)

monnier:
> I thought about that. I didn't add it because the response buffer is never
> edited, only replaced.  (`font-lock-extra-managed-props` is only for handling
> edits, not initial fontifying.) But if you think it's better to add them,
> too, I can do it. (There is definitely not harm in it.)

You might be right that it's not strictly needed in that case.
Please do include a comment about it, tho.

dominique-unruh:
@monnier I updated with commit with an explanatory comment.

Matafou:
> `describe-variable` on `font-lock-extra-managed-props` says that this variable becomes buffer local when set. So I guess this is OK like this.
> That's odd. AFAIK it is not "automatically buffer local". Maybe you have loaded a package which makes it "automatically buffer local" (that would be a nasty thing to do, but packages have been known to do worse).

I found the culprit: [highlight-indent-guides-mode](https://github.com/DarthFennec/highlight-indent-guides) does this [here](https://github.com/DarthFennec/highlight-indent-guides/blob/cf352c85cd15dd18aa096ba9d9ab9b7ab493e8f6/highlight-indent-guides.el#L971):
```lisp
(make-variable-buffer-local 'font-lock-extra-managed-props)
```

Matafou:
Thanks @dominique-unruh and @monnier!

PR #679, opened 2022-12-15
Fixed an error when `proof-ignore-for-undo-count` is a function.

rssoc:
As per `proof-ignore-for-undo-count`'s [declared type](https://github.com/ProofGeneral/PG/blob/bbbb1e2604996513cd05f54da1b8ef059a8e8968/generic/proof-config.el#L484-L490), it is possible for it to be a function.

This is a simple commit that fixes an error when you try using it as such.


Matafou (approved):
LGTM.

Issue #689, opened 2023-02-26
whitelist for admissible commands inside proofs


hendriktews:
In the last PG maintainer meeting we discussed the following solution for #688 and #687: Treat the current proof as a non-opaque proof if it contains a command starting with a capital letter, except for a number of white-listed harmless commands.
Non-opaque proofs (such as those terminated by Defined) are not
omitted.
@Matafou , @erikmd : Can you suggest commands for such a whitelist?

Matafou:
I will try to build a list when I have some time.
The things I see right now are 

Print Check About Search...
Time
Show

I see a problem with Module names: you can call a tactic by its long name: like Foo.tac. So you should probably ignoring commands starting with capitals if followed by a dot...

erikmd:
@hendriktews I didn't skim the Coq refman right now,
but I'd suggest to check the list of commands recognized by `C-c C-v` as state-preserving. Maybe there's a strong overlap here.

erikmd:
Yes I know. What I meant is that: thanks to this command, there is already a list of all commands that are state-preserving. Certainly not exhaustive w.r.t. the whitelist mentioned in #689. But in order to comply as much as possible with the principle Dont-Repeat-Yourself, I would suggest to just double-check manually what are these state-preserving commands, and maybe just add *programmatically in Lisp* a few more commands that are not state-preserving but which should be included in the whitelist. Typically, `Time`.

FTR, this list is defined here:

https://github.com/ProofGeneral/PG/blob/8416875696cb0c4283e96fe721d343277882ecea/coq/coq-syntax.el#L1245-L1247

and doing <kbd>C-h v coq-keywords-state-preserving-commands RET</kbd> yields:

<details>

```
("\\_<Obligations\\_>" "\\_<About\\_>" "\\_<Check\\_>" "\\_<Compute\\_>"
"\\_<fail\\_>" "\\_<Inspect\\_>" "\\_<Locate\\s-+File\\_>"
"\\_<Locate\\s-+Library\\_>" "\\_<Locate\\s-+Notation\\_>" "\\_<Locate\\_>"
"\\_<Print\\s-+Coercions\\_>" "\\_<Print\\s-+Hint\\_>" "\\_<Print Ltac2\\_>"
"\\_<Print\\_>" "\\_<Pwd\\_>" "\\_<Search\\_>" "\\_<SearchPattern\\_>"
[...]

PR #694, opened 2023-03-19
omit-proofs: handle commands that may have global effects

### LINKED ISSUE #688: "Omit proofs" breaks other proofs because it also skips hints
The proof of `bar` breaks when using "[Omit proofs](https://proofgeneral.github.io/doc/master/userman/Coq-Proof-General/#Omitting-proofs-for-speed)"  on `foo`:

```coq
Parameter P : nat -> Prop.
Axiom P_Z : P 0.
Axiom P_S : forall {n}, P n -> P (S n).

Lemma foo : P 1.
Proof.
  #[local] Hint Resolve P_Z P_S : core.
  (* ^^ This hint is skipped by PG when skipping [foo]'s proof. *)
  eauto.
Qed.

Lemma bar : P 3.
Proof.
  eauto. (* XXX Will fail when [foo] is skipped. *)
Qed.
```

This is with an _old_ proof general, but it doesn't look like much relevant has changed since then:
```
  proof-general                  20220803.1702  dependency            A generic Emacs interface for proof assistants
```

Ideally, PG would either support this script, warn about occurrences, or document such problems.

If you're wondering "why would you even consider writing this code", at least it can arise temporarily during development, and uses can be hard to find in larger scripts.

hendriktews:
Add the predicate `proof-script-cmd-prevents-proof-omission' to the omit-proofs framework, whose purpose is to check whether commands inside proofs might have global effects such that the proof must not be omitted.

Fixes #688


coq/coq.el
```diff
@@ -743,6 +743,18 @@ If locked span already has a state number, then do nothing. Also updates
...
+This predicate decides whether a command inside a proof might
+have effects outside the proof, which would prohibit omitting the
+proof, see `proof-script-omit-proofs'.
+
+Commands starting lower case are deemed as tactics that have
+proof local effect only. Everything else is checked against the
+STATECH field in the coq syntax data base, see coq-db.el."
+  (if (proof-string-match coq-lowercase-command-regexp cmd)
+      nil
+    (not (coq-state-preserving-p cmd))))
```
erikmd, line 757: as suggested in issue https://github.com/ProofGeneral/PG/issues/689#issuecomment-1476649322

I guess the lowercase test is unneeded; why not using something like this?

```suggestion
  (or (proof-string-match coq-keywords-state-changing-commands cmd)
    ; (not (proof-string-match coq-keywords-local-state-changing-commands cmd)) ; TODO
  ))
```

Regarding the TODO line, I rely on @hendriktews's suggestion to add another STATECH value and extend this:

https://github.com/ProofGeneral/PG/blob/8416875696cb0c4283e96fe721d343277882ecea/coq/coq-syntax.el#L1158-L1168

erikmd:
Sorry I believe some improvement of you for the omit-proof feature (maybe this PR, I didn't re-checked) was stalled because of belated review from Pierre and I; sorry for that.

So @hendriktews, do you think you'd need do you need further test or discussion from us on this PR?

Anyway I'd like to propose we pause our recurring PG maint telco this month (too many deadlines for me about grading or so in the upcoming two weeks), and resume in Feb…

Kind regards

hendriktews:
There has been no progress on the preferred solutions that we discussed last year. Therefore I propose merging this PR as is (assuming all tests pass). Erik, note that your suggestion does not work as proposed, because  `coq-keywords-state-changing-commands` is a list. As long as the proposed `coq-keywords-local-state-changing-commands` does not exist, it has the same problem with `Focus`, namely not omitting proofs that contain `Focus`.

hendriktews:
Hi @erikmd , no problem! I am pretty confident about this PR and about #696 in the sense that they are not perfect, but good enough. So I don't need anything. I just want to give you the chance to comment, but if you are busy, I'll merge in a week or so. I commit to improve this fix when someone is motivated to add the `coq-keywords-local-state-changing-commands` that we discussed.

PR #695, opened 2023-03-19
test-omit-proofs: skip for Coq 8.8 or older

hendriktews:
Test omit-proofs-never-omit-hints uses a feature ('#[local]') that has only been introduced in Coq 8.9. If we continue our current support policy, we will stop to support Coq 8.10 and older in April 2023. Instead of a proper fix, I therefore decided to simply skip the test for Coq 8.8 or older.

### DIFF
--- a/ci/simple-tests/omit_test.v
+++ b/ci/simple-tests/omit_test.v
@@ -41,15 +41,6 @@ Qed.
 
 (* automatic test marker 4 *)
 
-Lemma never_omit_hints : 1 + 1 = 2.
-Proof using.
-  #[local] Hint Resolve classic_excluded_middle : core.
-  (* automatic test marker 5 *)
-  auto.
-Qed.
-
-(* automatic test marker 6 *)
-
 Section let_test.
 
   Let never_omit_let : 1 + 1 = 2.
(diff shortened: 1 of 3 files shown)


erikmd:
Hi @hendriktews ! thanks for your PR ;

I just skimmed the code, but I got some doubt reading your remark

> If we continue our current support policy, we will stop to support Coq 8.10 and older in April 2023.

→ do you know where PG's support policy for Coq versions is documented?

Because AFAIK for Coq versions, no end-of-support policy is documented (hence the implied assumption that all Coq versions downto 8.4 or so are supported; thanks to the few `coq--version<` occurrences in the codebase !-)

hendriktews:
> The `#[local]` is indeed new but it is a generalisation of the older Local prefix. So maybe we can make this work for older versions?

I thought about this, but recognizing a command behind `#[local]` has its own challenges, therefore I would like to keep it. One could of course add a separate test that is compatible with 8.8 or so.

Matafou:
We had a discussion today with @hendriktews. This commit disables a *test*, and i only disables it when run with an old version of coq. It does not make any regression on pg's compatibility with coq versions. So it is completely OK to merge imho.

Issue #698, opened 2023-04-08
test_wholefile.v incompatible with 8.17


hendriktews:
The `Hint Resolve` command in `ci/test_wholefile.v` triggers a fatal warning in Coq 8.17. Test `060_coq-test-wholefile` in `ci/coq-tests.el` fails therefore for 8.17.

SnarkBoojum:
@hendriktews I've been trying to get in touch with you - especially about the Debian packaging for proofgeneral... could you contact me? I'm jpuydt and a Debian developer so you can guess a valid mail.

Thanks!

PR #699, opened 2023-04-08
ci/coq-tests: expect 060_coq-test-wholefile to fail for 8.17

hendriktews:
See also #698.


erikmd (approved):
Looks very fine to me, thanks @hendriktews

PR #708, opened 2023-10-03
Fix "Bad bounding indices" error on invalid _CoqProject

### LINKED ISSUE #707: confusing error on incorrect _CoqProject file
### The error

I have an incorrect _CoqProject file with the following content:

```
-R Category
```

When running Emacs+PG on a .v file in the same directory, the "Next step" or "Use buffer" commands error with an obscure message:

```
Bad bounding indices: 0, 2
```

Repro script:

```shell
$ mkdir repro
$ cd repro
$ echo "-R Foo" > _CoqProject
$ echo "Check true." > test.v
$ emacs test.v
```

Versions:
- Emacs: 28.3
- PG 20230414.931 from Elpa

(I observed the failure on older versions before, but I tried to make the issue go away by upgrading.)

### How I traced the error to the _CoqProject

I googled "Bad bounding indices", there were no hits whatsoever on the Coq resources, but some Github issues on some totally unrelated Emacs modes that were related to the `cl-subseq` function.

I then grepped the PG sources, there is a single use of `cl-subseq`:

https://github.com/ProofGeneral/PG/blob/e7aad1df36375284c65b496a44d346b047c01773/coq/coq-system.el#L550-L559

It looks like the code is trying to be robust against failures, but for some reason (maybe the `unless-debug` part?) this attempt fails on my machine and PG is unusable with broken _CoqProject files -- with a very obscure error message.

gasche:
Fixes #707.

The first commit is purely the bugfix. (Note: I didn't understand the previous code, as I am no Elisp expert, maybe it is doing something really important besides producing very obscure failure modes?)

The second commit turns `warn` into `message`. When testing with only the first commit I did not manage to see the `warn` string anywhere on my Emacs, so the error is a bit too silent to my taste. (We are noticing that the user _CoqProject is broken and just silently ignoring it.) `message` is slightly better. I wonder if `error` wouldn't be even better, but the current code logic is to be resilient and ignore the error, so I kept that.

### DIFF
--- a/coq/coq-system.el
+++ b/coq/coq-system.el
@@ -553,7 +553,7 @@ If ARITY is nil, return SWITCH."
   (cond
    ((not arity) switch)
    ((< (length raw-args) arity)
-    (warn "Invalid _CoqProject: not enough arguments for %S" switch)
+    (message "Invalid _CoqProject: not enough arguments for %S" switch)
     switch)
    (t
     (let ((arguments (cl-subseq raw-args 0 arity)))
--- a/coq/coq-system.el
+++ b/coq/coq-system.el
@@ -550,13 +550,14 @@ alreadyopen is t if buffer already existed."
 (defun coq--read-one-option-from-project-file (switch arity raw-args)
   "Cons SWITCH with ARITY arguments from RAW-ARGS.
 If ARITY is nil, return SWITCH."
-  (if arity
-      (let ((arguments
-             (condition-case-unless-debug nil
-                 (cl-subseq raw-args 0 arity)
-               (warn "Invalid _CoqProject: not enough arguments for %S" switch))))
-        (cons switch arguments))
-    switch))
+  (cond
+   ((not arity) switch)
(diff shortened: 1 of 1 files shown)


gasche:
cc @hendriktews : I don't know who to ping on this issue, but it is really simple and fixes a problem which cost me a bit of time

monnier:
The error in the original code is silly: `condition-case` has the following syntax `(condition-case VAR EXP1 (CONDITION EXP2))` where CONDITION says what errors we're catching.  But the code puts directly `EXP2` where `(CONDITION EXP2)` is expected, so `warn` ends up treated as CONDITION (a category of errors that doesn't actually exist).

Regarding `warn` vs `message`, `warn` is supposed to be *less* discrete than `message`, so I'm wondering what's going on.  Maybe it gets displayed in a buffer that gets immediately "unshown" because we're displaying some PG buffer instead in that same window.  This part should probably be turned into an Emacs bug.

Scutua (approved):

PR #712, opened 2023-10-28
ci/doc: document test strategy and the Coq/Emacs release table

hendriktews:
This is a first step. More documentation about the program `cipg` and `cipg` will follow.


hendriktews:
there was no objection, merging now

PR #714, opened 2023-10-31
qrhl-test: check translation of \sub

hendriktews:


hendriktews:
@dominique-unruh : this is just a proposal, it is up to you to decide what should be tested for qRHL. If you don't like this approach, no problem, then I'll abandon this PR.

PR #717, opened 2023-12-06
coq/coq-mode.el: Change the definition of 'coq--parent-mode

ayanamists:
## Problem Description

The current definition of 'coq--parent-mode is as follows:

```elisp
(defun coq--parent-mode ()
  (if coq-use-pg (proof-mode) (prog-mode)))
```

It currently functions correctly. However, this definition breaks the major mode hierarchy. For example:

```elisp
(get 'coq-mode 'derived-mode-parent)
; => 'coq--parent-mode
``` 

And the built-in Emacs function, `derived-mode-p`, also cannot recognize that `coq-mode` is a sub-mode of `prog-mode`:

```elisp
;; when you are in coq buffer
(derived-mode-p 'prog-mode)
; => nil
```

Some Emacs code, such as Spacemacs, relies on this information to determine if a buffer is a text/program file. For example, Spacemacs will display line numbers for all sub-modes of `prog-mode` and `text-mode` when `dotspacemacs-line-numbers` is set to `t` (See the [document](https://develop.spacemacs.org/doc/DOCUMENTATION.html#global-line-numbers)). However, due to this problem, Spacemacs does not display line numbers for `coq-mode` by default.

## Solution

It is better to define `'coq--parent-mode` as an alias:

```elisp
(defalias 'coq--parent-mode
  (if coq-use-pg 'proof-mode 'prog-mode))
```

With this definition, `derived-mode-p` can detect that `coq-mode` is a sub-mode of `prog-mode`:

```elisp
;; after editing
;; when you are in coq buffer
(derived-mode-p 'prog-mode)
; => 'prog-mode
```

And Spacemacs will automatically display line numbers for coq buffer when setting `dotspacemacs-line-numbers` to `t`.

### DIFF
--- a/coq/coq-mode.el
+++ b/coq/coq-mode.el
@@ -179,8 +179,8 @@ Near here means PT is either inside or just aside of a comment."
 ;;;###autoload
 (add-to-list 'auto-mode-alist '("\\.v\\'" . coq-mode))
 
-(defun coq--parent-mode ()
-  (if coq-use-pg (proof-mode) (prog-mode)))
+(defalias 'coq--parent-mode
+  (if coq-use-pg 'proof-mode 'prog-mode))
 
 ;;;###autoload
 (define-derived-mode coq-mode coq--parent-mode "Coq"


erikmd (approved):
Thanks a lot @ayanamists

There indeed seemed to be an issue over there…

Nitpick (I'm not sure here): should't we use sharp-quoting here?

namely, `#'proof-mode`, and so on.

@hendriktews / @Matafou, some opinion?

Matafou (approved):
Well spotted.

Matafou:
Thanks for spotting this. I am not sure there are a lot of people using the `coq-use-pg nil` . @ayanamists do you use it? Does it work? I think it is currently broken.

ayanamists:
> Thanks for spotting this. I am not sure there are a lot of people using the `coq-use-pg nil` . @ayanamists do you use it? Does it work? I think it is currently broken.

I have never used `coq-use-pg nil`. In fact, I don't understand why there's a need for such toggle.

Matafou:
Yes sorry I misread the code. The problem occurs when coq-use-pg is nonnil.

coq-use-pg allows some users that don't use PG for scripting but still want to an emacs mode for coq files. It deactivates the whole "prover interaction" part of proofgeneral and still provide syntax highlighting, indentation etc.

What I don't understand is how `coq--parent-mode` works. I think that proof-mode is itself derived from prog-mode. Isn't `derived-mode-p` a transitive test so that a parent of a parent is a parent?

monnier:
> coq-use-pg allows some users that don't use PG for scripting but still want
> to an emacs mode for coq files. It deactivates the whole "prover
> interaction" part of proofgeneral and still provide syntax highlighting,
> indentation etc.

That's right.  Not only that but it (should) make the code work even in
the absence of the rest of Proof-General (you only need the
`proof-general/coq` subdirectory and none of the other files).

The intention is to make it easier to reuse that code with some new
implementation using, say, SerAPI.

> What I don't understand is how `coq--parent-mode` works. I think that
> proof-mode is itself derived from prog-mode. Isn't `derived-mode-p`
> a transitive test so that a parent of a parent is a parent?

That's right.  But we want the code to work even if `proof-mode`
doesn't exist.

Matafou:
Thanks @monnier, good point. I am not sure to understand how an alias solves the problem here. I guess we can accept the PR.

Independently we probably need a little work to make the 'light" option really sane.

monnier:
> Thanks @monnier, good point. I am not sure to understand how an alias solves
> the problem here.

It's "magic" because `derived-mode-p` treats aliases specially, whereas
it wouldn't look inside the definition of a function like `coq--parent-mode`.

> I guess we can accept the PR.

Sounds good to me.

Matafou:
Thanks again @ayanamists !

Scutua (approved):

PR #720, opened 2023-12-27
Fix typos

pitmonticone:


hendriktews:
Thanks!
(For the future I would suggest to consider a commit message "fix typos" or "fix typos in <file>" for such changes.)

PR #726, opened 2024-01-16
remove executable rights from .el files

### LINKED ISSUE #704: generic/pg-goals.el shouldn't be executable
I don't think there's a good reason for that?

hendriktews:
Fixes #704.


erikmd:
Good point, thanks Hendrik!

PR #737, opened 2024-02-13
Avoid evaluating void variables

haselwarter:
The last `unless` case of `unicode-tokens-copy-configuration-variables` tries to evaluate variables which may be undefined via `symbol-value`, which leads to an error. This commit adds a check to make sure the variables are only accessed if bound.

How this ever worked without errors I don't know...


Matafou:
Hi. thanks for the PR. Sorry for the delay...

Do we really want this? If one uses `unicode-tokens.el` then isn't it mandatory to have `token-symbol-map` defined? If yes then failing looks good to me so that the user knows what is missing. If no then we should either do what this PR does, or decide to default to something else.

@cpitclaudel and @hendriktews what do you think?

Matafou:
Any idea about this?

PR #742, opened 2024-03-02
coq: run silently and explicitly Show when necessary

hendriktews:
This is a step towards fixing #568. It fixes the cases after Proof, comments and auto, leaving now 3 instead of 6 failing tests in ci/simple-tests/test-goals-present.el.

The first commit here breaks prooftree. More work is needed.

### DIFF
--- a/ci/coq-tests.el
+++ b/ci/coq-tests.el
@@ -196,6 +196,8 @@ For example, COMMENT could be (*test-definition*)"
 
 
 (ert-deftest 020_coq-test-definition ()
+  ;; There are no infomsgr when running silent.
+  :expected-result :failed
   "Test *response* output after asserting a Definition."
   (coq-fixture-on-file
    (coq-test-full-path "test_stepwise.v")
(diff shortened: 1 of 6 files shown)


hendriktews:
Hi @erikmd , @Matafou , @cpitclaudel ,
AFAICT, the difficult cases of PG not showing or showing incorrect goals are solved now. IIRC, the conclusion in #467 was that PG should run Coq completely in silent mode. This PR precisely does this. If you have an opinion on whether we should prefer the approach of this PR over the alternatives in #467 or #429, please tell! If you have some time to try this out, it would be nice to hear if this PR breaks other stuff.
What still needs to be done is clearing the goals after Qed and Admitted, I believe this is simple. What is harder is to get proof-tree to work again.

hendriktews:
The approach used in this PR does not work for Search and Check commands. I'll start from scratch in a new PR.

cpitclaudel:
Thanks a lot for working on this @hendriktews ; I don't have anything smart to say on the implementation at this stage but I'm very glad that you're on it :)

PR #743, opened 2024-03-03
coq-tests: fix error introduced by Cyril Anaclet in 0ae25c5a

hendriktews:
The purpose of unwind-protect is to execute the cleanup-forms in case of a non-local exit. It does not make any sense to move these forms behind unwind-protect.


monnier:
Agreed: looks like a paren-typo.

erikmd:
Good catch :100: thank you Hendriktews !!

PR #745, opened 2024-03-08
CI: do not artificially restrict parallel test execution

hendriktews:
For whatever reason, the test job always runs late in a CI run (I am blaming the max-parallel annotation for this). This leads to longer CI runs, because, at the end, for several rounds on 6 instances of the test job are executing. Note that Github starts at most 20 of our tests in parallel, so it is not our responsibility to ensure that limit.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,9 +38,6 @@ jobs:
           - 29.1
           - 29.2
           # CIPG change marker end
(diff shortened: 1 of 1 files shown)


erikmd (approved):
Thanks @hendriktews ! indeed I don't see any drawback to raise the max number of these parallel jobs.

PR #750, opened 2024-03-25
poor PG man's workaround for coq/coq#11479

hendriktews:
Work towards a feature that permits to extract lists of failing proofs from PG. Similar to what coq/coq#11479 would provide.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -14,6 +14,9 @@ the Git ChangeLog, the GitHub repo https://github.com/ProofGeneral/PG
     Proof-tree visualization is currently only supported for Coq. The
     prooftree support has been substantially rewritten, making use of
     the much better support since Coq version 8.11.
+*** New command `proof-check-proofs' to generates the proof status
+    of all opaque proofs. Currently only available for Coq, see Coq
+    changes below for more details.
 
 ** Coq changes
 *** support Coq 8.19
@@ -23,15 +26,21 @@ the Git ChangeLog, the GitHub repo https://github.com/ProofGeneral/PG
      modules as coqdep error.
 *** Renew support for proof-tree visualization, see description in
     generic changes above.
-
-
-**** New options coq-compile-extra-coqc-arguments and
-     coq-compile-extra-coqdep-arguments to configure additional
-     command line arguments to calls of, respetively, coqc and coqdep
-     during auto compilation.
-
-**** Fix issues #687 and #688 where the omit-proofs feature causes
-     errors on correct code.
+*** New command `proof-check-proofs' to generates the proof status
+    of all opaque proofs. This command is useful for a development
+    process where invalid proofs are permitted and vos compilation and
+    the omit proofs feature are used to work at the most interesting
+    or challenging point instead of on the first invalid proof. The
+    command generates a list of all opaque proofs in the current
+    buffer together with the information whether the proof script is
+    currently valid or invalid. The command can also be run in batch
+    mode, for instance in a continuous integration environment.
+*** New options coq-compile-extra-coqc-arguments and
+    coq-compile-extra-coqdep-arguments to configure additional
+    command line arguments to calls of, respetively, coqc and coqdep
+    during auto compilation.
+*** Fix issues #687 and #688 where the omit-proofs feature causes
+    errors on correct code.
 
 
 * Changes of Proof General 4.5 from Proof General 4.4
--- a/generic/pg-user.el
+++ b/generic/pg-user.el
@@ -600,6 +600,186 @@ last use time, to discourage saving these into the users database."
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;
+;; Check validity of proofs
+;;
+
+(defun proof-check-report (proof-results)
+  "Report `proof-check-proofs' results in PROOF-RESULTS in special buffer.
+Report the results of `proof-check-proofs' in buffer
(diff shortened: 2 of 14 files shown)


Matafou (commented):
LGTM. Working like a charm. The "XXX" in docstrings should probably fixed, although for most of them the name of the functions/constants are quite self-explanatory.

hendriktews:
This is only an intermediate state, it needs some cleanup. I intent to add a batch mode feature, where the result are written into a file, maybe in the test anything protocol (TAP). And I also want to do tests and documentation.
Maybe the whole patch should go into `pg-user.el`, do you have an opinion about this?

Matafou:
I think it would the right place (if if the code is sufficiently prover-independent of course).
TAP seems a good idea too.

hendriktews:
Hi @erikmd and @Matafou ! I consider this mostly finished. I still plan to add a section in the adapting manual and a changelog entry. Do you want to discuss this PR before merging?

erikmd:
Thanks a lot Hendrik!

I did not have the time this week, but I plan to test it on next Tuesday!

Cheers

hendriktews:
The last force-push switched to TAP version 13 (instead of 14). I was told 14 has issues and therefore not as wide tool support.

CHANGES
```diff
@@ -23,15 +26,21 @@ the Git ChangeLog, the GitHub repo https://github.com/ProofGeneral/PG
...
-     errors on correct code.
+*** New command `proof-check-proofs' to generates the proof status
+    of all opaque proofs. This command is useful for a development
+    process where invalid proofs are permitted and vos compilation and
+    the omit proofs feature are used to work at the most interesting
+    or challenging point instead of on the first invalid proof. The
+    command generates a list of all opaque proofs in the current
+    buffer together with the information whether the proof script is
+    currently valid or invalid. The command can also be run in batch
+    mode, for instance in a continuous integration environment.
```
erikmd, line 37: CI is a good use case as well indeed. Did you already tried this? I mean, not from an internal PG elisp test, but from an example GHA script that would call emacs like this? `
```
emacs --batch --eval "(progn (setq user-init-file \"~/.emacs\"))" --load ~/.emacs --eval "(message \"etc.\")"
```

And maybe I'm mistaken, but to facilitate this, maybe the function `proof-check-proofs` should query from some way the **filename** of the file to test? (which could be the current buffer in interactive mode, for example)

generic/pg-user.el
```diff
@@ -597,6 +597,223 @@ last use time, to discourage saving these into the users database."
...
+    ;; determine a relative file name for current buffer
+    (when buffer-file-name
+      (setq coq-proj-dir (locate-dominating-file buffer-file-name
+                                                 coq-project-filename)))
+    (if coq-proj-dir
+        (setq src-file (file-relative-name buffer-file-name coq-proj-dir))
+      (setq src-file (buffer-name)))
+
+    ;; generate header
+    (with-current-buffer (get-buffer-create proof-check-report-buffer)
```
erikmd, line 629: Some feedback:
* (important) I believe the `*proof-check-report*` should be read-only, adding `(read-only-mode)` somewhere here
* (minor, maybe to do later, not in this PR!) It could be nice to just type `RET` on one of the lemma name, to directly switch to it (handy if the .v file is very large), you know, a bit like the `*ert*` view of `M-x ert`

erikmd (commented):
Dear @hendriktews, sorry for replying late! I've just tested the feature.

It is very nice and useful, thanks! in particular as a regular user of PG's omit-proofs feature :)

I wrote some feedback in the review comments below.

Cheers

erikmd:
> The last force-push switched to TAP version 13 (instead of 14). I was told 14 has issues and therefore not as wide tool support.

OK looks good, but I'm not TAP-savvy, so I can't comment more on this one.

CHANGES
hendriktews: No GHA and no other repo but I manually tested it a lot from the shell. Please see the PG manual for the command line. Adding a test invoking `proof-check-proofs` from a makefile is probably a good idea.

generic/pg-user.el
hendriktews: * read-only: good point
* key binding: good idea, but I am not motivated to do this now. Instead I envision an extension of `proof-check-proofs` that annotates failing theorems in the Coq sources with `(* FAIL *)`. For my repos I then want to set up a CI action that only checks the consistency of these `FAIL` comments using `proof-check-proofs`. This way, fixed proofs and new failing ones pop up prominently during review and one can search form them. Moreover, `M-.` (`xref-find-definitions`) works in the report buffer.

hendriktews:
> OK looks good, but I'm not TAP-savvy

Neither me, but our local QA expert is.

generic/pg-user.el
erikmd: > annotates failing theorems in the Coq sources with `(* FAIL *)`

OK, looks like a nice idea!

> `M-.` works in the report buffer.

Great!

CHANGES
hendriktews: I investigated the error yesterday: 3-pane mode (on by default) is broken if the frame height is too small and certain Emacses set the frame height to 9 inside a docker container.
In the commit I added yesterday, the test you asked for runs as part of the simple tests. Do you have the opinion that this is not enough?

erikmd: > certain Emacses set the frame height to 9 inside a docker container.

OK!

> In the commit I added yesterday, the test you asked for runs as part of the simple tests. Do you have the opinion that this is not enough?

No worries! it is enough as is.

Thanks a lot @hendriktews !

PR #751, opened 2024-03-28
DON'T MERGE: investigate doc-magic problem in PR #750

hendriktews:


hendriktews:
The problem was caused by Stefan, pushing directly, without running CI

Issue #605, opened 2021-10-06
Feature request: use goal count information printed at top of goals buffer in modeline


cpitclaudel:
I won't have time to implement this, but it should be easy and it should provide a nice quality of life improvement.

Each goal buffer stars with a header line, like this:

```
1 focused subgoal
(shelved: 1) (ID 339)
  
  width : Z
  BW : Bitwidth width
  …
```

It would be nice to move that info to the modeline, next to the buffer name (`*goals*`).  We can change the "lighter" of the response mode to be a variable for that, and then update the variable when we update the response buffer.

This would save a bit of space, which is always nice with large Coq goals.

Matafou:
As you suggest we could show it in the goals buffer modeline of course. I suppose it is more intuitive.

Issue #168, opened 2017-03-08
`Abort`-ing doesn't clear the goal window (or show the previous current goal)


cpitclaudel:
@Matafou Is this another case of the problems we noticed recently? Thanks!

Matafou:
Probably.

PR #755, opened 2024-03-31
Add features: use goal count information printed at top of goals buffer in modeline, clear the goal window when proof is skipped (abort, admitted...).

axe1d:
In response to issues #605  and #168, I moved the information of current goal to the *goals* buffer modeline (to save some space and visibility basically), and cleared the *goals* buffer when encountering skip proof tactics like Abort. or Admitted. in Coq.

### DIFF
--- a/generic/pg-goals.el
+++ b/generic/pg-goals.el
@@ -107,8 +107,10 @@ so the response buffer should not be cleared."
       (bufhist-checkpoint-and-erase))
 
     ;; Only display if string is non-empty.
-    (unless (string-equal string "")
-      (funcall pg-insert-text-function string))
+    (when (not (string-equal string ""))
+      (funcall pg-insert-text-function string)
+      (when (string-equal proof-assistant "Coq")
+        (coq-parse-goals-info string)))
 
     (setq buffer-read-only t)
     (set-buffer-modified-p nil)
--- a/generic/proof-script.el
+++ b/generic/proof-script.el
@@ -1486,6 +1486,8 @@ Besides stuff that is not yet documented here, this function
   and sections.
 - enters some commands and their spans in some database (with for
   me unknown purpose)"
+  (when (string-equal proof-assistant "Coq")
+    (reset-goals-modeline)) ; Erase goal modeline when a proof is closed
   (unless (or (eq proof-shell-proof-completed 1)
 	      ;; (eq proof-assistant-symbol 'isar)
 	      )
--- a/coq/coq.el
+++ b/coq/coq.el
@@ -1261,6 +1261,62 @@ width is synchronized by coq (?!)."
   :safe 'booleanp
   :group 'coq)
 
+;; update *goals* buffer modeline
+(defun coq-parse-goals-info (goals-output)
+  "Parse output to extract and update goals information in modeline."
+  (let ((goals-count-regexp "^\\([0-9]+\\) goals? (ID \\([0-9]+\\))")
+        (num-goals 0)
+        (current-goal-id nil))
+    (when (string-match goals-count-regexp goals-output)
+      (setq num-goals (string-to-number (match-string 1 goals-output)))
+      (setq current-goal-id (match-string 2 goals-output)))
+    (update-goals-modeline num-goals current-goal-id)
+    ; hide goals info on first line
+    (with-current-buffer proof-goals-buffer
+      (goto-char (point-min))
+      (let ((line-end (line-end-position)))
+        (let ((overlay (make-overlay (point-min) line-end)))
+          ; invisible line
+          (overlay-put overlay 'invisible t))))))
+
+(defun update-goals-modeline (goal id)
+  "Extracts proof state and updates the goals buffer modeline dynamically."
+  (with-current-buffer proof-goals-buffer
+    (let ((goal-str (if (> goal 1) "goals" "goal")))
+      (setq-local my-goals-info
+                  (format "%d %s, ID: %s" goal goal-str id)))
+      (let ((baseline-mode-line-format
+             '("%e"
+               mode-line-front-space
+               mode-line-mule-info
+               mode-line-client
+               mode-line-modified
+               mode-line-remote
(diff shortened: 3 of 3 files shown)


Matafou:
The change is nice. A few remarks here and there. 2 questions:

- should we remove the code that displays this information in the scritpting buffer then?
- the `coq-test-coqtop-unavailable.el` fix seems unrelated but real. @hendriktews, @erikmd should we fix this?

ci/simple-tests/coq-test-coqtop-unavailable.el
```diff
@@ -18,7 +18,7 @@
   (setq coq-prog-name "unavailable-program")
 
   ;; ensure coq-prog-name cannot be found
-  (should (not (locate-file coq-prog-name exec-path)))
```
hendriktews: Could you explain this change?

etc/testsuite/.#simple.v
hendriktews, line 1: What is this file for?

hendriktews:
>     * the `coq-test-coqtop-unavailable.el` fix seems unrelated but real. @hendriktews, @erikmd should we fix this?

The error is related to this PR, it was introduced in the first commit f6a425f.

Matafou:
The PR should first be cleaned up (useless files, then reduced to one commit imho.

ci/simple-tests/coq-test-coqtop-unavailable.el
axe1d: As I mentioned earlier, I wanted to appropriate this project by making sure my building was fine, so I broke a simple test first. But now I think my PR should be a bit cleaner.

axe1d:
> The PR should first be cleaned up (useless files, then reduced to one commit imho.

I thank you all for taking the time for reviewing my PR (which is my first ever, that's why it was a bit messy, and I apologize for that). I think it is a bit cleaner, and I welcome any advice on how to improve it before it can be merged (if it's needed of course).

generic/pg-goals.el
```diff
@@ -79,6 +79,63 @@ May enable proof-by-pointing or similar features.
 ;;
 ;; Goals buffer processing
 ;;
+
+;; update *goals* buffer modeline
+(defun coq-parse-goals-info (goals-output)
```
hendriktews, line 84: PG is a generic interface for proof assistants, see https://proofgeneral.github.io/doc/master/userman/Introducing-Proof-General/#Introducing-Proof-General, which does not only support Coq, see https://proofgeneral.github.io/doc/master/userman/Introducing-Proof-General/#Supported-proof-assistants . Assistant specific code should therefore go into the assistant specific directory.

generic/pg-goals.el
```diff
@@ -107,8 +164,9 @@ so the response buffer should not be cleared."
       (bufhist-checkpoint-and-erase))
 
     ;; Only display if string is non-empty.
-    (unless (string-equal string "")
-      (funcall pg-insert-text-function string))
+    (when (not (string-equal string ""))
+      (funcall pg-insert-text-function string)
+      (coq-parse-goals-info string))
```
hendriktews, line 169: I am pretty sure that calling a Coq specific function here will break any other proof assistant.

generic/proof-script.el
```diff
@@ -1486,6 +1486,7 @@ Besides stuff that is not yet documented here, this function
   and sections.
 - enters some commands and their spans in some database (with for
   me unknown purpose)"
+  (reset-goals-modeline) ; Erase goal buffer when a proof is closed
```
hendriktews, line 1489: The comment here is inconsistent with your documentation of reset-goals-modeline.

hendriktews:
I have the impression that the "clear goal window" part that is mentioned in the title of this PR is not present in the patch anymore. Can you confirm this? (I also suspect that this aspect did not work in your previous version, at least not for Admitted, as I don't remember that you have changed that the `goals-reset-after-admitted` is expected to fail, as it was before I merged PR #748 -- but I guess this suspicion is completely irrelevant now.)

generic/pg-goals.el
hendriktews: @Matafou Can you recomment where this code should be added in the `coq` subdir?

generic/proof-script.el
hendriktews: Maybe this call does not break other assistants, but it should not be called for all assistants. Please check if you can find a suitable hook to add this call.

hendriktews:
I am not convinced every user will be fond of this change. The mode line is usually crowded and often it is truncated and some parts are not visible. Maybe this change should therefore be customizable, @erikmd @Matafou what is your opinion here?

Matafou:
My opinion is that everything should be customizable, but I don't always apply it to myself. I think the current goal counter (on the scripting buffer) is not customizable either.

axe1d:
Hi everyone, I modified my feature for Coq-only implementation, moved my functions to the coq subdir, and adapted the modeline update only when PG is instantiated for Coq. My goal here is to offer the possibility to users to custom the modeline as mentioned in #605. I thank you all again for this precise review of my code.

Matafou:
I have been thinking about this PR a bit more and I think we should not merge it.

1. This feature (having the number of goals in the modeline) already exists: the number of goals is already printed in the modeline of the script buffer. If we want to move it to the goals buffer we should reuse and modify/clean/optionalize the existing code.

See https://github.com/ProofGeneral/PG/blob/734bcdb27fdcf54fbe6ab05bf957408e70c69748/coq/coq.el#L3121 and following, in particular the line:

```lisp
(add-hook 'proof-shell-handle-delayed-output-hook #'coq-update-minor-mode-alist)
```

Feel free to submit a new PR adapting this part of the code.

2. The "clear buffer" stuff should go in a separate PR.

Remark on the code: Testing the name of the prover in the generic code is not the way PG is supposed to be configured. Specific provers should not appear in the generic part.

If noone object I shll close this PR soon.

Issue #760, opened 2024-04-16
3-pane mode broken with small frame heights


hendriktews:
In the default setting, PG splits the frame into 3 windows regardless of the frame height. If the frame height is less than 12, then this signals a window too small for splitting error. This is a real problem for CI where emacs 26.3, 27.1, and 27.2 set their default frame height to 9 when running inside a container.

PR #761, opened 2024-04-16
fix 3-pane mode for small frame heights

hendriktews:
These tests are for #760 and some of them fail because of this problem.


hendriktews:
@erikmd , @Matafou : If there are no objections I am going to merge this and soon afterwards #750.

PR #762, opened 2024-04-17
Coq: run silently and explicitly Show when necessary - second attempt

hendriktews:
This is another attempt towards fixing #568.


hendriktews:
@Matafou , @erikmd , @cpitclaudel : With the option `coq-run-completely-silent` to switch back to old behavior, I believe the risk of breaking a use case of some user is manageable. Therefore I suggest to merge this in about a week.

erikmd:
Hi @hendriktews!

(I just quickly skimmed the PR code but) LGTM, thanks.

> With the option ..., I believe the risk of breaking a use case of some user is manageable. Therefore I suggest to merge this in about a week.

Agree

Matafou:
Maybe I need to test how it interacts with debug mode. I don't have much time this week though.

Matafou:
@hendriktews can you tell us what remains to be tested/adapted pls? I think this PR should not rot.
You said elsewhere that the prefix arg C-u for queries like coq-Print wouldn't work. What else?

Matafou:
> I use this code on a daily basis, so it is working for me. Remaining things that I know:
> 
>     * coq-Print with prefix arg
> 
>     * prooftree - I have local fixes on top of the commits in this PR to get prooftree working, but they need to be polished before uploading
>
>     * I have never tried diff or debug mode, it would be nice, if somebody who knows what to expect could try this

I can test this.

>     * Not sure if this is a problem: [...] This might make the impression, search does not work at all.

I think this is a problem. I must admit I am always in 3 windows mode.

hendriktews:
> > * Not sure if this is a problem: [...] This might make the impression, search does not work at all.
> 
> I think this is a problem. I must admit I am always in 3 windows mode.

So it is not a problem for you. It is only a problem when there is no window showing the response buffer when you issue the command. Note that any nonempty response from the search command is correctly shown.

Meanwhile I discovered another problem:

* When you are in proof mode without goals, i.e., just before you type QED, then no response from a search or locate command is visible, even in case these commands yield results.

Matafou:
> > > * Not sure if this is a problem: [...] This might make the impression, search does not work at all.
> > 
> > 
> > I think this is a problem. I must admit I am always in 3 windows mode.
> 
> So it is not a problem for you. It is only a problem when there is no window showing the response buffer when you issue the command. Note that any nonempty response from the search command is correctly shown.

No it is not a problem for me, but the lack of user feedback in this case is a (small) problem.

> Meanwhile I discovered another problem:
> 
>     * When you are in proof mode without goals, i.e., just before you type QED, then no response from a search or locate command is visible, even in case these commands yield results.

Indeed. It also happens after other messages like "This subproof is complete, ...".

The reason is that the search command is now followed by a `Show`. In this case the Show command triggers a response "No more goals.", which is detected as another response message instead of a goal message. Hence it immediately overwrites the response to the query.

I tested and adding  "No more goals." to  `proof-shell-start-goals-regexp` solves this problem, but it seems to trigger other problems, e.g. the message "No more goals." is not displayed in goals as expected.  I think it is because "No more goals." is already in  `proof-shell-clear-goals-regexp` and `proof-shell-proof-completed-regexp`. Understanding the interaction between these 3 variables is needed I guess.

Matafou:
I support merging this now.
And see what happens with `C-u` prefix, debug mode (probably both rarely used) and anything else interacting with this. If ever we get an avalanche of bug reports then we will just switch the default value of `coq-run-completely-silent` to nil.

hendriktews:
OK, let's wait for another few days, in case somebody else wants to try it.
I have a simple fix for `C-c C-a C-a` with empty response, that also fixes `C-c C-a C-a` just before `Qed`. The same idea should also work for `C-u coq-Print`, I just need to roll it out.

hendriktews:
Please excuse the delay, I thought it would be good to first add tests for the searches that get broken by this PR. Hopefully they are now ready, see #827

ci/coq-tests.el
```diff
@@ -413,6 +422,13 @@ For example, COMMENT could be (*test-definition*)"
 ;; (coq-should-buffer-regexp (regexp-quote "The command has indeed failed with message: Tactic failure: Cannot solve this goal.") "*response*")
 
 (ert-deftest 091_coq-test-regression-Fail()
+  ;; XXX What is the difference between this test and
```
Matafou, line 425: The point in coq file is not the same: "(*FailNoTrace*)" vs "(*FailTrace*)". Concretelty the difference is the command `Local Unset Ltac Backtrace.`.

ci/coq-tests.el
```diff
@@ -200,6 +200,8 @@ For example, COMMENT could be (*test-definition*)"
 
 
 (ert-deftest 020_coq-test-definition ()
+  ;; There are no infomsgr when running silent.
+  :expected-result :failed
```
Matafou, line 204: We should probably test this when only one command is sent at a time.

Matafou:
This works like a charm. Thanks a lot @hendriktews for this work.

ci/coq-tests.el
hendriktews: OK, it would be nice if the documentation in that file would describe what is tested there.

ci/coq-tests.el
hendriktews: AFAICT there is only one command processed in this test. Even if you process one definition alone, there is no infomsg when running silent.

Matafou: You are right. this behaviour is ok for me. I wonder if someone really care about this fonfirmation messages outside pure textual coqtop use.

Matafou:
After a few days of work the only annoying thing for me is the "Set some option; do the command; unset the option" triggered by the C-u prefix. Like `C-u C-a C-b eq RET`. The commans are sent, but the last  unset command makes the response buffer immediately empty.

hendriktews:
Do you mean `C-u C-c C-a C-b`, i.e., `coq-About`? Have you tried PR #831, which is supposed to fix these cases? (See also the tests `response-buffer-visible-coq-check-print-all-inside-poof` and `response-buffer-visible-coq-check-print-all-poof-end` in `ci/simple-tests/coq-test-goals-present.el`)
The problem that is still left is when you type `About eq.` in your proof script buffer just before a `Qed` after you saw the `No more goals` message.

PR #767, opened 2024-04-30
DONT MERGE - test strange CI behavior

hendriktews:


hendriktews:
I used this to analyze the error I got in #765. Turns out the passing combinations use git 2.20 (maybe earlier), which does not seem to have a problem if the .git directory is owned by a different user and not writable. The failing ones use git 2.30 (or newer), which does not recognize the .git directory during a git diff in such a situation.

PR #768, opened 2024-05-05
Reduce splash time to 1s.

Matafou:
The idea is that this splash screen annoys everyone, even in its new manga-style form.

I would even consider removing it completely but at least let us not have it last for 8 seconds by default.

### DIFF
--- a/generic/proof-splash.el
+++ b/generic/proof-splash.el
@@ -42,7 +42,7 @@
   :type 'boolean
   :group 'proof-user-options)
 
-(defcustom proof-splash-time 8
+(defcustom proof-splash-time 1
   "Minimum number of seconds to display splash screen for.
 The splash screen may be displayed for a wee while longer than
 this, depending on how long it takes the machine to initialise


hendriktews:
I have `(proof-splash-enable nil)` since the second day I use PG. So yes, please merge.
(IMO it would be better to have it 3 or 4 seconds together with a bit button "Disable this splash screen for the future" at the front, but I don't feel inclined to spend time on it.)

erikmd:
AFAIAC, I only press `q` and it disappears immediately. Maybe this shortcut could just be made more visible?
or improving this splash screen so that yet another shortcut is put forth, to dismiss the screen once and for all?

erikmd:
YMMV anyway, and I don't object to merging this

erikmd:
@Matafou Ready to merge this now?

Even if Hendrik suggested a possible enhancement

> (IMO it would be better to have it 3 or 4 seconds together with a bit button "Disable this splash screen for the future" at the front, but I don't feel inclined to spend time on it.)

it seems that the current version of this PR has the merit to be simple, and still greatly improve the situation

Matafou:
Sure. Sooner is better imho.

PR #772, opened 2024-06-11
also omit proofs with bullets and braces

hendriktews:
Currently bullets and braces are classified as commands that prevent proof omission, which is wrong.

### DIFF
--- a/ci/simple-tests/coq-test-omit-proofs.el
+++ b/ci/simple-tests/coq-test-omit-proofs.el
@@ -258,7 +258,6 @@ This test only checks the faces in the middle of the proof."
   (should (eq (first-overlay-face) 'proof-omitted-proof-face)))
 
 (ert-deftest omit-proofs-omit-bullets-and-braces ()
-  :expected-result :failed
   (let ((proof-omit-proofs-option t)
         pos-10)
     (message "omit-proofs-omit-bullets-and-braces: Check bullets and braces")
--- a/ci/simple-tests/coq-test-omit-proofs.el
+++ b/ci/simple-tests/coq-test-omit-proofs.el
@@ -106,7 +106,7 @@ In particular, test that with proof-omit-proofs-option configured:
 
   ;; Check 1: check that the proof is valid and omit can be disabled
   (message "1: check that the proof is valid and omit can be disabled")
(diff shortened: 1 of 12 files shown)


erikmd (approved):
Very good point, thanks @hendriktews !

I proofread the PR and everything looks good to me.

erikmd:
Hi @hendriktews !
sorry for the nitpick, but I believe that you use the Merge/Rebase button (?), which has 2 drawbacks documented in
https://github.com/ProofGeneral/PG/wiki/Checklist-for-testing-and-reviewing-a-PR#merging-a-pr

so AFAICT, it'd be preferable we either use Merge/Sqash or "regular" Merge (for PRs with several commits) next time.

Kind regards

hendriktews:
yes, I prefer merge/rebase because of the linear history. Personally, I would not consider the mentioned drawbacks as a problem. We don't sign commits, the only problem remains are the PR numbers. Do you consider the PR numbers valuable?

erikmd:
Hi @hendriktews, thanks for your comment.

> We don't sign commits

yes and no → we don't have a policy to commits manually, but github does automatically sign commits for any squash or regular merge, which is an (admittedly small) upside

> do you consider the PR numbers valuable?

a bit (because it may refer to the related discussion in PR comments) but I can agree that this is subjective!

> because of the linear history

Yes, the merge/rebase will produce a completely linear history. What are the upsides of this git tree from your opinion?

(just try to understand all the arguments so that we may agree it's better to either "never do a regular merge", or "always do a regular merge or squash-merge")

erikmd:
BTW, this makes me think that there might be a small risk when doing a merge/rebase:

if upstream/master has extra commits not part of the pr-branch (still without creating a conflict),

the merge/rebase creates commits that have a totally different SHA1 and git tree,
so that it is possible that none of the intermediate commits (before the tip of upstream/master after the merge-rebase) builds successfully, even if all commits of the pr-branch were fine.

(but admittedly, this is a minor risk as the most important thing is that the tip of master if fine)

erikmd:
again, sorry for the nitpick! but I'm happy to discuss whether I should do things differently (or not), from your viewpoint.

PR #780, opened 2024-07-08
Fixes #779 regression cannot step Fail correctly.

### LINKED ISSUE #779: REGRESSION: ProofGeneral cannot step over Fail correctly
This problem was originally reported as #597 and fixed in 2021.
Now with Coq 8.20+rc1, the message for `Fail` is
```
The command has indeed failed with message:
Tactic failure: Cannot solve this goal.
In nested Ltac calls to "now (tactic)" and "easy", last call failed.
```
And PG does not step over `FAIL`.
To reproduce, see `ci/test_stepwise.v`, assert to the end of the line containing `FailTrace`, and assert the next line.
(Before 8.20 the message printed by coq seemed to be
```
The command has indeed failed with message:
In nested Ltac calls to "now (tactic)" and "easy", last call failed.
Tactic failure: Cannot solve this goal.
```
)

Matafou:
As suggested by @hendriktews: added a "no-error" regexp.

Also adapted the tests so that they pass with coq pre-8.20 and post-8.20. No need to ass more tests case imho.

### DIFF
--- a/coq/coq-syntax.el
+++ b/coq/coq-syntax.el
@@ -1296,19 +1296,6 @@ Very similar to `coq-omit-proof-admit-command', but without the dot."
 
 (defvar coq-symbols-regexp (regexp-opt coq-symbols))
 
-;; HACKISH: This string matches standard error regexp UNLESS there is
-;; the standard header of the "Fail" command (which is "The command
-;; blah has indeed failed with message:\n"). The case where the error
-;; header has nothing before it is treated using "empty string at
-;; start" regexp. BUT coq-error-regexp (and hence
-;; proof-shell-error-regexp) must be correct either when searching in
-;; a string or when searching in the proof-shell-buffer when point is
-;; at the start of the last output. Hence when we use \\` (empty
-;; string at start of the string) we should also accept \\= (empty
-;; string at point).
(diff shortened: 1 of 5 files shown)


Matafou:
Initial issue report: #779 .

coq/coq-syntax.el
```diff
@@ -1321,13 +1321,10 @@ Very similar to `coq-omit-proof-admit-command', but without the dot."
     "\\<Error:")
   "A list of regexps matching coq error headers.")
 
-(defvar coq--raw-error-regexp (coq--regexp-alt-list coq--error-header-re-list))
+(defvar coq-error-regexp (coq--regexp-alt-list coq--error-header-re-list))
```
hendriktews, line 1324: `coq--prefix-not-regexp` is not needed any  more, is it? Do you want to delete it together with the comment in front of it?

Matafou: Forgot that. will do.

hendriktews:
Thanks for this PR, it now looks fine to me. I suggest to merge when the tests are successful.
At some stage we should add `proof-shell-no-error-regexp` to PG-adapting.

PR #774, opened 2024-06-15
fix(coq.el): (setq proof-shell-strip-crs-from-input nil)

erikmd:
Related: https://github.com/ProofGeneral/PG/issues/773

@hendriktews do you agree that we merge this ? (if the CI passes, of course)

PR #782, opened 2024-07-09
Fix #781 PG does not position to error.

### LINKED ISSUE #781: PG does not position to error
This bug appeared because of PR #774. Backtracking it is not satisfying because #773 shows a good reason for it.

Example:
```coq
Require Import Utf8.
Definition trois := 3. (*test-definition*)
Definition β := 3. (*test-definition*)
Definition β' := 3. (*test-definition*)
Definition β'' := 3. (*test-definition*)
Print trois.
Eval compute in (λ γγ:nat, γγ)
                       4 +
                       (λ a:nat, a) 4
                     + ααα * trois * trois.
```
the error is not positioned on `ααα`. This has nothing to do with utf8, but any fix should check that it works ok for this.

Matafou:
We cannot use the " ^^^ " thing to compute the error location because commands are not stripped of newlines anymore (#774).

Solution: use the character position information with the following subtleties

- position are given by coq as **bytes** positions.
- position given by coq is the position from the previous "." fed to coq. I.e. we must be careful to strip any blank after the final "." ofthe previous (hence from all) commands.

### DIFF
--- a/ci/coq-tests.el
+++ b/ci/coq-tests.el
@@ -255,12 +255,83 @@ For example, COMMENT could be (*test-definition*)"
    (coq-test-full-path "test_stepwise.v")
    (lambda ()
      (coq-test-goto-before " (*test-lemma2*)")
-     (let ((proof-point (save-excursion (coq-test-goto-after "(*error*)")))) 
-     (proof-goto-point)
-     (proof-shell-wait)
-     (coq-should-buffer-string "Error: Unable to unify \"false\" with \"true\".")
-     (should (equal (proof-queue-or-locked-end) proof-point))))))
+     ;; redefining this function locally so that self removing spans
+     ;; remain longer. Cf span.el
+     (cl-letf (((symbol-function 'span-make-self-removing-span)
+                (lambda (beg end &rest props)
+                  (let ((ol (span-make beg end)))
+                    (while props
+                      (overlay-put ol (car props) (cadr props))
+                      (setq props (cddr props)))
+                    (add-timeout 10 'delete-overlay ol)
+                    ol))))
+       
+       (let ((proof-point (save-excursion (coq-test-goto-after "(*error*)") (point)))
+             (proof-cmd-point (save-excursion
+                                (coq-test-goto-after "(*error*)")
+                                (re-search-forward "reflexivity")
+                                (re-search-backward "reflexivity")))) 
+         (proof-goto-point)
+         (proof-shell-wait)
+         (coq-should-buffer-string "Error: Unable to unify \"false\" with \"true\".")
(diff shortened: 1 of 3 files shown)


Matafou:
This needs a bit of a test and discussion.

It *seems* to work but it is difficult to be sure.

Things to check that I can think of.

- Is the computation of position robust to any encoding?
- Currently I strip trailing spaces of a command otherwise coq's error location are wrong. Should make this an option (in case it is not suitable for other provers)?

Matafou:
I just added a test that checks that the correct region is highlighted.
@hendrik or @erikmd can you review pls? I sum up the fix:

Note: testing this I found something [strange in error location info from coq](https://github.com/coq/coq/issues/19355)

Matafou:
This is not ready yet. I will try to fix this by the end of the summer.

Matafou:
The two failing tests are on coq-8.11. I will try to investigate but maybe this is not blocking.
One problem remains: support the coq bug that has just been fixed in coq master?

Matafou:
This seems ready. I will merge this tomorrow if no one objects.
@erikmd this is an important fix imho, is there something to do to update pg distributed packages?

Matafou:
This PR

- fixes #781 
- makes use of the character location given by coq instead of the "^^^^"
- revealed https://github.com/coq/coq/issues/19355 in coq wrt to error location of curly braces.
-  works around it by detecting suspicious locations (probably almost 100% accurate) and defaulting to something reasonable (beginning of command).
- introduces 2 tests of error highlighting including one specific to the workaround.
- has not been tested very widely, I guess users will complain if something is wrong.

Matafou:
- re-introduced removal of error location from the response buffer.
- understood and partially fixed an old small bug on error locations: coq gives wrong location if one insert spaces at the end of a command (before the newline). We used to insert "`Set/Unset Silent. `" with a trailing space. I leave to another PR the removal of other similar wrong strings. I only fix this one because it happens all the time.

Matafou:
I decided to add the fix of hardwired command strings (removing trailing spaces) in a separate commit  of this same PR.

erikmd:
Thanks a lot @Matafou !

sorry for replying late

> @erikmd this is an important fix imho, is there something to do to update pg distributed packages?

It depends on the distribution you think of:

* for MELPA, as soon as the PR has been merged in PG master, it is rebuilt later on during the same day
* for NonGNU-devel ELPA, this is the same
* for NonGNU ELPA (stable releases), we will need to do a release — note that a tag is not enough, see this comment by Stefan: https://github.com/ProofGeneral/opam-switch-mode/pull/10#issuecomment-1599495287

PR #787, opened 2024-09-07
CI: add Coq 8.20

hendriktews:
Testing for Emacs 29.1 is and will be further reduced, because Ubuntu 23 reached it's end of life. Some containers for Emacs 27.1 will soon be deleted, because Debian Bullseye reached it's end of life.


Matafou (approved):
LGTM.

Scutua (approved):

PR #789, opened 2024-09-10
Fix #757 indentation of "\in"

### LINKED ISSUE #757: Indentation issue : "\in"
SSReflect uses `\in` as opposed to the unicode `∈` to indicate membership in lists. However, the former is not indented correctly.
Example:
```
Lemma test:
  forall a : nat,
    a \in [::] ->
          False. (* indented by 4 spaces*)
```
as opposed to:
```
Lemma test:
  forall a : nat,
    a ∈ [::] ->
    False. (* not indented *)
```

Matafou:
The lexer now glues a "\" to an immediately following word if it is itself preceded by a space.

Note that for really good indentation you should try to add something like this to your configuration:

(setq coq-smie-user-tokens '(("\\in" . "=")))

to tell the indentation grammar that \in has the same precedence as "=".

### DIFF
--- a/ci/test-indent/indent-backslash-ops.v
+++ b/ci/test-indent/indent-backslash-ops.v
@@ -1,0 +1,23 @@
+Require Import mathcomp.ssreflect.ssrbool.
+
+Definition xx := nat.
+Module foo.
+  (* from PG issue #757 *)
+  Lemma test:
+    forall a : nat,
+      a \in [::] ->  (* "\in" should be detected as one token *)
+      False.
+  Proof.
+  Abort.
+  Qed.
+
+  Lemma test2:
+    forall a : nat,
+      a \in  (* "\in" should be detected as one token *)
+        [::] ->
+      False.
+  Proof.
+  Abort.
+  Qed.
+End foo.
+
--- a/coq/coq-smie.el
+++ b/coq/coq-smie.el
@@ -266,14 +266,18 @@ the token of \".\" is simply \".\"."
 
 
 ;; A variant of smie-default-backward-token that recognize "." and ";"
-;; as single token even if glued at the end of another symbols.
-
+;; as single token even if glued at the end of another symbols. We
+;; glue "\" in front of a word though, to follow ssreflects
+;; ocnvention.
 (defun coq-backward-token-fast-nogluing-dot-friends ()
   (forward-comment (- (point)))
   (let* ((pt (point))
          (tok-punc (skip-syntax-backward "."))
          (str-punc (buffer-substring-no-properties pt (point))))
-    (if (zerop tok-punc) (skip-syntax-backward "w_'"))
+    ;; skip a regular word + one backslash
+    (when (zerop tok-punc)
+      (skip-syntax-backward "w_'")
+      (if (looking-back "\\s-\\\\") (forward-char -1)))
     ;; Special case: if the symbols found end by "." or ";",
     ;; then consider this last letter alone as a token
     (when (and (not (zerop tok-punc)) (string-match "\\s.+[.;]" str-punc))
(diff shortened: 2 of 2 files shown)


Matafou:
The problem was that \in was lexed as "\" "in" and that "in" is a (pretty crowdedly overloaded) reserved keyword for let in, Let in,..

I will wait a feedback from #757 before merging this.

ci/test-indent/test_backslash_ops.v
```diff
@@ -0,0 +1,19 @@
+Require Import mathcomp.ssreflect.ssrbool.
```
erikmd, line 1:
```suggestion
From mathcomp Require Import all_ssreflect.
```

ci/test-indent/test_backslash_ops.v
```diff
@@ -0,0 +1,19 @@
...
+      False.
+  Proof.
+  Abort.
+
+  Lemma test2:
+    forall a : nat,
+      a \in  (* "\in" should be detected as one token *)
+        [::] ->
+      False.
+End foo.
```
erikmd, line 18:
```suggestion
  Proof.
  Abort.
End foo.
```

erikmd (approved):
Hi @Matafou, thanks a lot for this fix!

I tested it successfully on another file of mine, as well as on your test file and it worked-out-of-the-box™ (using Emacs 28.2 on Debian 12)

ci/test-indent/test_backslash_ops.v
Matafou: right!

Matafou:
I don't know the exact use of `\xxx` in mathcomp, are all these `\symbols` non associative infix operators? 

Note that without the `coq-smie-user-tokens` trick these tokens will be considered with a default rule, precedence and associativity (@monnier can you tell what are these for a token not appearing in the rules and grammar, starting with a `\`). This can lead to strange indentation maybe. @vzaliva tell us if it is the case it should be easy to fix.

erikmd:
> I don't know the exact use of `\xxx` in mathcomp,

Basically, the default notation convention in the mathcomp core library is to *only use ASCII* operators, not UTF-8 or so,
and typically uses notations very close to those from LaTeX. E.g.:
* `x \notin E`
* `f \o g \o h`
* `\sum_(0 <= i < n) (i + 1)`
* etc., see also https://www-sop.inria.fr/teams/marelle/MC-2022/cheatsheet.pdf

> are all these \symbols non associative

not exactly, cf. the `\o` operator for function composition.

> infix operators?

You also have `\sum_i`, `\prod_i`, `\bigcap_i`, `\bigcup_i` and a few others.

Matafou:
OK, I was hoping for something simple. How naive I am :-).

If we can come up with a decent `coq-smie-user-tokens` declaration for usual ssreflect operators we could put them into `coq-standard-token-synonyms`. 

It would look like:

```lisp
(setq coq-smie-user-tokens '(("\in" . "=") ("\notin" . "=") ("\o" . "+") ... )) 
```
(I guess `\sum()` is lexed/parsed as a standard function application and that's ok).

Maybe this already exists somewhere?

Matafou:
Yes the fix is for any token starting with `\`. But indentation is not the same for a infix associative operator and a prefix one.

erikmd:
OK thanks!

So IIUC, in addition to this PR, it just suffices to provide a relevant `(setq coq-smie-user-tokens '(…))` list so indenting ssreflect would work out-of-the-box?

BTW I tested `(setq coq-smie-user-tokens '(("\in" . "="))` vs. `(setq coq-smie-user-tokens '(("\in" . "+"))` and didn't notice a specific change in the indentation; could you help by providing a minimal complete example where the two infixes are indented differently?

Matafou:
The difference can be subtle (and this is very fragile, but here is a difference between `+` and `=` (that would reflect on `\in`):

``` coq
Lemma foo:
  forall x y,
    y
    =
      z
    = y.

Lemma foo:
  forall x y,
    y
    =
      z
      + y.
```

Matafou:
Note that this is a bit wrong since `=` is not associative, but that is the idea.

erikmd:
OK thanks @Matafou !

So I'll try to come up with a comprehensive list of the \in | \notin | … SSR symbols this WE.

Question: can/should this PR be merged beforehand? or implementing the `coq-standard-token-synonyms` list you mentioned will require a non-trivial refactoring this PR's code?

Matafou:
I think we can merge.

PR #793, opened 2024-10-02
proof-shell: Don't ask about killing the proof assistant on exit

hendriktews:
This PR is only about the commit in the title. To avoid conflicts, it builds on PR #791.

I have disabled this silly question about killing Coq when quitting Emacs decades ago already. IMO we don't need a user option to control this behavior.

### DIFF
--- a/generic/proof-shell.el
+++ b/generic/proof-shell.el
@@ -479,6 +479,12 @@ process command."
 	(erase-buffer)
 	(proof-shell-set-text-representation)
 
+        ;; If the user quits Emacs together with the Proof General
+        ;; session, it's clear that the proof assistant needs to be
+        ;; killed - don't ask.
+        (set-process-query-on-exit-flag
+         (get-buffer-process proof-shell-buffer) nil)
+
 	;; Initialise associated buffers
 	(with-current-buffer proof-response-buffer
 	  (erase-buffer)
--- a/CHANGES
+++ b/CHANGES
@@ -19,6 +19,11 @@ the Git ChangeLog, the GitHub repo https://github.com/ProofGeneral/PG
     changes below for more details.
 *** New command `proof-check-annotate' to annotate all failing proofs
     with FAIL comments.
+*** Improve splash screen, add menu entry to permanently disable it
+    (Proof-General -> Quick Options -> Display -> Disable Splash Screen),
+    reduce splash screen time to make it less annoying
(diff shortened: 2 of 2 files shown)


generic/proof-shell.el
```diff
@@ -477,6 +477,11 @@ process command."
 	(erase-buffer)
 	(proof-shell-set-text-representation)
 
+        ;; If one quits Proof General, it's clear that the proof
+        ;; assistant needs to be killed - don't ask.
```
erikmd, line 481: On the one hand I agree with your remark;
but on the other hand I believe changing would be not "uniform" w.r.t. other emacs features?

e.g. if we do `M-x shell RET`
then `C-x C-s`, exactly as you say, "if we quit emacs, it's clear that the nested shell needs to be killed", but emacs will still ask for confirmation.

Would you agree to customize this?

monnier:
>> +        ;; If one quits Proof General, it's clear that the proof
>> +        ;; assistant needs to be killed - don't ask.

[ Nitpick: the flag affects the behavior when one quits Emacs, rather
  than Proof General.  I don't think we even have a clear notion of
  "quitting Proof General".  ]

> e.g. if we do `M-x shell RET`
> then `C-x C-s`, exactly as you say, "if we quit Emacs, it's clear that the
> nested shell needs to be killed", but Emacs will still ask for confirmation.

I'm not sure which behavior is better here (my own use of Coq is
sufficiently limited that I don't think it's representative), but the
main issue is whether killing the process would lose information, IOW
it's whether the state of the process is likely to contain information
that the user cannot trivially recreate (e.g. by restarting the process).

For `M-x shell` we consider that it's sufficiently common to have such
important state (for example because the shell is currently running
a command whose interruption would leave a mess).

erikmd:
@monnier thanks for your feedback!

regarding your main point:

> it's whether the state of the process is likely to contain information
> that the user cannot trivially recreate (e.g. by restarting the process).

it appears than unlike `M-x shell` processes, coqtop processes generally generally don't contain "critical state information", so I guess this amounts to a +1 for @hendriktews 's proposition.

andreas-roehler:
Being in favour of this change, however, optional. Changing used defaults commonly comes with hidden costs.

hendriktews:
Rebased to master, reformulated "quit PG" in response to Stefan.

hendriktews:
accidentally merged the CHANGES note from #791, removing this now

PR #812, opened 2025-02-24
Makefile: Replace "which" by POSIX "command -v"

ulm:
Forwarding [Gentoo Linux bug 950240](https://bugs.gentoo.org/950240).

Gentoo is replacing the `which` command in Makefiles and other build scripts, because it is not portable, and its existence isn't guaranteed by POSIX.

Debian does the same since some time, see https://lwn.net/Articles/874049/ for an elaborate rationale.

This PR replaces `which` by `command -v` which is mandated by POSIX.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@
 
 # Set this according to your version of Emacs.
 # NB: this is also used to set default install path names below.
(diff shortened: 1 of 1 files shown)


hendriktews:
Thanks a lot for this patch! Running CI now.

PR #813, opened 2025-02-28
Fix low level code to find the start of a command.

Matafou:
A small bug in the code for finding the real start of a command. It was wrong in the case of focus commands `3:{`.

The side effects of this should be almost invisible. I detected the bug when debugging experimental stuff.

### DIFF
--- a/ci/test-indent/indent-tac.v
+++ b/ci/test-indent/indent-tac.v
@@ -269,6 +269,19 @@ Module M1.
         { auto.
         }
       }
+      1:{
+        destruct n.
+        2:{ auto.
+        }
+        { auto. }
+      }
+      {
+        destruct n.
+        2:{
+          auto. }
+        { auto.
+        }
+      }
     Qed.
   End M2.
 End M1.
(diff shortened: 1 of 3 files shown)


Matafou:
This tests error are strange. Given the error message I doubt the problem comes from my changes. 

@hendriktews @erikmd, any clue? Most of them are those:

```
Pull docker image
  COQ_IMAGE=proofgeneral/coq-emacs:coq-8.20-rc1-emacs-26.3
  
  Error response from daemon: No such image: proofgeneral/coq-emacs:coq-8.20-rc1-emacs-26.3
  Error response from daemon: manifest for proofgeneral/coq-emacs:coq-8.20-rc1-emacs-26.3 not found: manifest unknown: manifest unknown
```

and the one on coq-8.11 has this:

```
Test test-coq-par-job-needs-compilation-quick condition:
      (cl-assertion-failed
       ((> same-counter 0)
        "create files with same time stamp filed"))
  Error:    FAILED  1/1  test-coq-par-job-needs-compilation-quick
  ```

Matafou:
re-launching failed tests. Just in case.

Issue #790, opened 2024-09-30
PG does not position to error of vernacular commands


hendriktews:
When I process a buffer just containing
```
Comput x.
```
I get `error in process filter: End of buffer`, because PG tries to move point to an illegal position. With some non-empty lines below `Comput`, PG puts point to a wrong location. With just newlines below `Comput`, I get some other error.

Same for `Require xxx.`, where `Require` is valid, but module `xxx` does not exist.

When I switch back to 0e0170f96f5feaeefe59052a08373080acc20393, before #774 error positioning does work fine.

hendriktews:
fixed by #816

Issue #814, opened 2025-03-20
Delay when first command produces an error


SkySkimmer:
On a file such as 
~~~coq
Check Tr.
~~~
PG takes about 1s to report the error ("the reference Tr was not found in the current environment").

hendriktews:
Thanks for the report. What Emacs and Coq/Rocq version are you using? What command do you use to process that line?
(I get the error immediately with Emacs 28.2, Coq 8.17.1 and `proof-assert-next-command-interactive`.)

SkySkimmer:
GNU Emacs 28.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.38, cairo version 1.16.0) of 2025-02-27, modified by Debian
(debian stable, I get the same issue with the debian testing emacs)
PG e0ec3db200f8ed465ec533b6409e64e3dfb6a9b0
Tested with Coq 8.11, Coq 8.20.1, Rocq master

PR #816, opened 2025-03-21
Fix problems with error on the first command of a file.

Matafou:
Small annoying bug seemingly squashed.

I think this solves #814 and the fact that the error was not correctly located.


hendriktews:
This PR does also fix #790.

PR #820, opened 2025-04-01
Fix make magic

hendriktews:
Pick two commits from Stefan from March 2024 that fix `make magic` for 30.1 and also improve it. I reverted these commits back then in a race and misunderstanding in order to regain an error-free CI baseline on master.

### DIFF
--- a/doc/PG-adapting.texi
+++ b/doc/PG-adapting.texi
@@ -3090,12 +3090,12 @@ To set the default value for these settings in prover-specific cases,
 you should use the special @code{defpgdefault} macro:
 
 @c TEXI DOCSTRING MAGIC: defpgdefault
-@deffn Macro defpgdefault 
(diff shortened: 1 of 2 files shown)


Matafou:
LGTM. I don't really understand what changed in 30.1. Anyway it works and it is cleaner.

monnier:
What changed is the representation of interpreted functions.
I.e. the evaluation of `(lambda ...)` used to return something of
the form `(closure ...)`, i.e. a list.  Now it returns an object of its
own type (very similar to the objects returned when the code is
byte-compiled).

So any ELisp code which looks inside function objects is affected, such
as:

```
         (argsyms   (cond ((eq (car-safe def) 'lambda)
			   (nth 1 def))
                          ((eq (car-safe def) 'closure)
                           (nth 2 def))))
```

PR #827, opened 2025-05-11
CI: add new tests for coq-Search and friends

hendriktews:
The tests check that coq-Search and similar functions correctly show their response in two-pane mode, even if the response is empty.


hendriktews:
If nobody objects, I'll merge this tomorrow.

PR #831, opened 2025-05-15
Fix response

hendriktews:
Fix display of the response buffer for display and query commands
generated by Proof General when Coq runs completely silent for the
following corner cases:
- when the response is empty
- in the no more goals state
- when printing all is temporarily switched on for the display/query
  command


Matafou:
Is there any reason to wait? The corner case of a query issued when a "no more subgoals" deserves a fix, but I think our users can live without it. A fortiori since we merged #762.

hendriktews:
no, there is no reason any more AFAICT

PR #841, opened 2025-08-12
Rocq: prefer Rocq binaries over Coq ones

hendriktews:
Before this change Proof General ignores an Opam Roqc installation if there is a system wide Coq installation.


hendriktews:
@Matafou Did you explicitly choose to prefer Coq binaries over Rocq ones?

Matafou (approved):

Issue #845, opened 2025-09-15
Emacs 31 warns on missing lexical bindings for `generic/proof-autoloads.el`


iwahbe:
Since this is the auto-loads file, this warning shows up when I start Emacs. The warning can be addressed by adding a lexical binding cookie to the file in question. To preserve existing behavior and silence the warning, we can use:

```patch
 generic/proof-autoloads.el | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/generic/proof-autoloads.el b/generic/proof-autoloads.el
index 8090aa0..4f35a9f 100644
--- a/generic/proof-autoloads.el
+++ b/generic/proof-autoloads.el
@@ -1,4 +1,4 @@
-;;; proof-autoloads.el --- automatically extracted autoloads
+;;; proof-autoloads.el --- automatically extracted autoloads  -*- lexical-binding: nil; -*-
 
 ;; This file is part of Proof General.
 
```

iwahbe:
@monnier That sounds reasonable. Do you know if the file is auto-generated? I'd be happy to open a PR if I knew where the correct root change was.

PR #846, opened 2025-09-23
Lexical autoloads

iwahbe:
This PR makes 2 changes:
- It adds `-*- lexical-binding: t; -*-` to `generic/proof-autoloads.el`. This is preserved by `make autoloads`, so no other change is needed to address #845.
- It adds an explicit `(require 'autoload)` to the command for `make autoloads`, since `autoload` is not required by default in Emacs 31 (and possibly earlier). 

Fixes https://github.com/ProofGeneral/PG/issues/845


Matafou:
Hi. Thanks for this PR. Some test fail:

     FAILED  test-coq-par-job-needs-compilation-quick

with message:

    create files with same time stamp failed

@hendriktews @monnier do you have an idea? Why would autoloads mess with this?

Matafou:
Indeed the rerun went well.
The goal of having lexical binding everywhere in PG is still uncertain, but I guess we can merge this one. @hendriktews do you agree?

PR #856, opened 2025-12-13
feat(coq): insert named goal selectors

dhalilov:
This PR implements the `coq-insert-named-goal-selectors` function that automatically generates named goal selectors (of the form `[name]: tac.`)  for each open named goal. I bound it to `C-c C-a C-g` by default.

The implementation is based on the output of `Show Existentials`, and notably checks for the absence of `(only printing)` in the output to distinguish focusable goal names from unfocusable goals (see https://github.com/rocq-prover/rocq/pull/20809).

This feature is mainly intended to be used in conjunction with [automatically named goals](https://rocq-prover.org/doc/master/refman/proofs/writing-proofs/proof-mode.html#rocq:flag.Generate-Goal-Names) in Rocq 9.2.

Here's a demo:
demo

### DIFF
--- a/coq/coq.el
+++ b/coq/coq.el
@@ -2924,18 +2924,7 @@ Also insert holes at insertion positions."
       (unless (proof-string-match coq-error-regexp match)
         (let ((start (point)))
           (insert match)
-          (indent-region start (point) nil)
-          (let ((n (holes-replace-string-by-holes-backward start)))
-            (pcase n
-	      (0 nil)                   ; no hole, stay here.
-	      (1
-	       (goto-char start)
-	       (holes-set-point-next-hole-destroy)) ; if only one hole, go to it.
-	      (_
-	       (goto-char start)
-	       (message
-                (substitute-command-keys
-                 "\\[holes-set-point-next-hole-destroy] to jump to active hole.  \\[holes-short-doc] to see holes doc."))))))))))
+          (holes-replace-string-by-holes-backward-jump start))))))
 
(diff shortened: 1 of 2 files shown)


rswarbrick (commented):
Hi there! I was feeling enthusiastic about reading some elisp again, so I thought I'd go through and leave a review. I hope it's useful!

rswarbrick:
This looks good to me. I did a re-review and realised that all the stuff that had made me raise my eyebrows is now deleted! :-D

Matafou:
Hi! Very nice idea! I want to merge but I have one question first.

I see that you sneaked into holes.el (which I thought nobody cared about). But I am planning to remove holes.el in a short future and rely on yasnippet instead (opt in). What do you think of that? I will adapt your code when this happens of course.

dhalilov:
> I see that you sneaked into holes.el (which I thought nobody cared about). But I am planning to remove holes.el in a short future and rely on yasnippet instead (opt in). What do you think of that?

I used holes.el since that was the solution used in `coq-insert-match`, but I would definitely be in favor of using yasnippet 😄

cpitclaudel:
> C-c C-a C-g by default.

One thing here: having a keybinding that ends in C-g is a bit of a no-no in Emacs land, because it breaks the usual meaning of C-g as "cancel the current action".  A few modes used to do it (including reST-mode), but they have stopped.

Perhaps C-c C-a M-RET as a parallel to C-c C-a RET?  Or perhaps C-c C-a C-c for "cases"?

PR #863, opened 2026-01-18
Fix compile warnings: Doc string single quote

MorganJamesSmith:
Hello again!

When compiling proof general there are over 400 warnings!  Some of these warnings might be important but it is hard to tell with this much noise.  I would like to help reduce this noise.

Newer emacs emit warnings when that are unescaped single quotes in docstrings.  I've spent a good chunk of time resolving this issue.

For symbols that link to a variable or function they should be written as: `symbol'

For symbols that should not link to anything as they are just a symbol they should be written as: \\+`symbol'.

By going through and making sure symbols are quoted using this style, I was also able to simplify the documentation generator slightly which removes some false positives and removes a lot of "@samp{@code{symbol}}" and replaces it with "@code{symbol}" which seems better.


Matafou:
Thanks!

PR #867, opened 2026-02-12
Add support for 'repeat-mode'

username231419:
Emacs has a `repeat-mode` that you can use to avoid typing in the prefix of a keychord for related commands.  Eg. when enabled you can just press `o` after `C-x o` to switch to then next window.  This adds similar support for at least two commands that I find useful when loading and undoing commands in a Rocq buffer.

### DIFF
--- a/coq/coq.el
+++ b/coq/coq.el
@@ -961,7 +961,7 @@ Otherwise propose identifier at point if any."
            (t (coq-id-or-notation-at-point)))))
     (read-string
      (if guess (concat s " (default " guess "): ") (concat s ": "))
-     nil 'proof-minibuffer-history guess)))
+     nil 'proof-minibuffer-history guess t)))
 
 
 (defun coq-ask-do (ask do &optional dontguess postformatcmd wait)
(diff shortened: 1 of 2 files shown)


coq/coq.el
```diff
@@ -961,7 +961,7 @@ Otherwise propose identifier at point if any."
            (t (coq-id-or-notation-at-point)))))
     (read-string
      (if guess (concat s " (default " guess "): ") (concat s ": "))
-     nil 'proof-minibuffer-history guess)))
+     nil 'proof-minibuffer-history guess t)))
```
Matafou, line 964: Is this related to the repeat-mode?

Matafou (commented):
This looks like a nice idea.

coq/coq.el
username231419: You are right, it is not.  This is just an unrelated change I had commuted and accidentally pushed in this PR.  Should I remove it, because I would have also wanted to submit it as well anyway.

Matafou: It is probably better to do another PR. But we can leave it here. What is the recommended policy for this?

PR #874, opened 2026-04-14
Allow EasyCrypt lemma names to begin with `_`.

alleystoughton:
Fix the regular expression `easycrypt-goal-command-regexp` to take out the `nosmt` option (which is no longer EasyCrypt syntax) and allow lemma names to begin with `_`. Note
that the end of this regular expression must only match the *beginning* of a lemma name,
not the whole name. (Without this fix, it was impossible to undo a lemma with such a name.)

### DIFF
--- a/easycrypt/easycrypt-syntax.el
+++ b/easycrypt/easycrypt-syntax.el
@@ -49,7 +49,7 @@
 
 (defconst easycrypt-goal-command-regexp
   (concat "^\\(?:local\\s-+\\)?\\(?:" (proof-ids-to-regexp easycrypt-keywords-proof-goal) "\\)"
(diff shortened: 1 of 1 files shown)


Matafou (approved):
LGTM.

Matafou:
Sorry to be so slow to answer. Feel free to ping me more next time :-).

alleystoughton:
> Sorry to be so slow to answer. Feel free to ping me more next time :-).

Thanks! Will this automatically show up in MELPA?

monnier:
> Thanks! Will this automatically show up in MELPA?

It should automatically show up in [NonGNU-devel
ELPA](https://elpa.nongnu.org/nongnu-devel/proof-general.html) and AFAIK
in [MELPA](https://melpa.org/#/proof-general) as well, yes.

PR #879, opened 2026-07-30
Fix #522: PG does not ignore comments in _CoqProject file.

### LINKED ISSUE #522: PG does not ignpre comments in _CoqProject file
_CoqProject file has a bash-like comment syntax. `#` starts a comment until the end of the line.

But if PG find a _CoqPrpjet file containing, e.g.
```
# -R . Foo
```
 it will then consider the commented -R directive:
```
 coqtop -topfile <scriptfile> -emacs -R <expanded path> Foo
```

Matafou:
Remove Lines starting with `#` are now ignored when reading a project file.


Matafou:
Susbsumed by #880.

Issue #392, opened 2018-09-27
Syntax of `-arg` in `_CoqProject`


herbelin:
Hi, I'm writing in relation with a CoqIDE issue coq/coq#5773.

The `_CoqProject` file supports a syntax `-arg` which is not documented in the Coq reference manual but which we fortunately found documented in ProofGeneral. We realized that there are however some ambiguities with this syntax, in particular regarding options which span over several "tokens" of the command line, as e.g. for `-arg "-w -notation-overriden"`, vs `-arg -w -notation-overriden`, vs one line `-arg -w` and the next line `-arg notation-overriden`, etc.

That would be good to be able to document more precisely this syntax and that PG, CoqIDE and `coq_makefile` parse it consistently. For instance, is `-arg` intended to take a space-free argument, or, if non space-free, a double-quote surrounded argument? Are single-quote surrounded argument allowed? How to pass an argument which contains quotes, as, if, e.g., I would like to pass `-arg '-Q "" Foo'`? Etc. In the case of `coq_makefile` the parsing rules are [here](https://github.com/coq/coq/blob/ecbd506624a1898bca658a8b1c4cfe13f1f9976c/lib/coqProject_file.ml#L98-L137). What are the parsing rules for PG?

PS: "Remark" is written using the French spelling [here](https://github.com/ProofGeneral/PG/blob/95786256fc32ad6881b4d95d98b5b41acd171fc9/doc/ProofGeneral.texi#L4385).

cpitclaudel:
Hi @herbelin!

The parsing code in PG is here:
https://github.com/ProofGeneral/PG/blob/733cd24a7368ee186884da488da0f59bbedb627e/coq/coq-system.el#L471-L537

Note in particular the implementation `coq--extract-prog-args`, which further splits the arguments passed to `-arg`.

You can use the following to test how it behaves:

```
(coq--read-options-from-project-file "…contents of _CoqProject…")
```

And you can use the following to see what we pass to Coq:

```
(coq--extract-prog-args (coq--read-options-from-project-file "..."))
```

For example: 

```
M-: (coq--extract-prog-args (coq--read-options-from-project-file "-arg \"-w -notation-overriden\"")) RET
⇒ ("-emacs" "-w" "-notation-overriden")
```

**Concretely**:

- `-arg "-w -notation-overriden"` is supported and yields 
- `-arg -w -notation-overriden` is not supported
- `-arg -w -arg notation-overriden` is supported
- Single quotes are not recognized 
[...]

PR #880, opened 2026-07-31
Adjust _RocqProject parsing to `rocq makefile` changes

### LINKED ISSUE #113: PG is unusable when `-w` is passed to Coq via _CoqProject
If my `_CoqProject` file contains

```
-arg "-w '-deprecated-appcontext -notation-overridden'"
```

then PG is unable to start coqtop.  It seems that it tries to pass `"-w" "'-deprecated-appcontext" "-notation-overridden'"` to `coqtop`, which is wrong.

cpitclaudel:
This is an attempt to follow `lib/coqProject_file.ml` instead of doing our own parsing:

- `#` starts a comment line;
- Single quotes toggle quoting and get stripped;
- `\` isn't an escape character, so `-arg "-w \"a b\""` (which we used to accept) now needs to be written `-arg "-w 'a b'"`.

Fixes #113, #392, #522, #589.  Closes #879.

### DIFF
--- a/ci/simple-tests/coq-test-coqproject-args.el
+++ b/ci/simple-tests/coq-test-coqproject-args.el
@@ -29,10 +29,8 @@
     ("-arg \"'a b' 'c d'\""                          ("a b" "c d"))
     ("-arg \"'a' 'b'\""                              ("a" "b"))
     ("-arg \"a'b c'd\""                              ("ab cd"))
-    ("-arg \"a 'b c' d 'e\""                         ("a" "b c" "d" "e"))
     ("-arg \"-Q '' Foo\""                            ("-Q" "" "Foo"))
     ("-arg \"''\""                                   (""))
-    ("-arg \"'\""                                    (""))
     ("-arg \"\""                                     nil)
     ("-arg \"   \""                                  nil)
     ("-arg \"-w\tall\""                              ("-w\tall"))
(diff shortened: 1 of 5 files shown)


RalfJung:
> \ isn't an escape character

Is this based on experimenting with `rocq makefile` / reading its code? I don't actually know and never tried what `\` does.

RalfJung:
This seems to work fine with the _RocqProject I was trying to use earlier. :)
Thanks a lot!

cpitclaudel:
> Is this based on experimenting with rocq makefile / reading its code? I don't actually know and never tried what \ does.

Reading the code.  It used to just dump whatever you added into Makefile, and that's how I wrote the original code.  Then that changed, and my -arg code broke.  So I rewrote the -arg code to track the new rocq makefile closely.

Matafou:
Thanks @cpitclaudel ! LGTM. The errors in the CI seem unrelated (problem with the generation of to files with same date stamp).

cpitclaudel:
Thanks!  I'll have a quick look at fixing the tests and then I'll merge

cpitclaudel:
@Matafou fixed the par tests; can you have a look?  I did my best at reconstructing what the tests meant ^^

ci/simple-tests/coq-test-coqproject-args.el
```diff
@@ -0,0 +1,68 @@
...
+    ("-arg \"'a b'\""                                ("a b"))
+    ("-arg \"'a b' c\""                              ("a b" "c"))
+    ("-arg \"c 'a b'\""                              ("c" "a b"))
+    ("-arg \"'a b' 'c d'\""                          ("a b" "c d"))
+    ("-arg \"'a' 'b'\""                              ("a" "b"))
+    ("-arg \"a'b c'd\""                              ("ab cd"))
+    ("-arg \"a 'b c' d 'e\""                         ("a" "b c" "d" "e"))
+    ("-arg \"-Q '' Foo\""                            ("-Q" "" "Foo"))
+    ("-arg \"''\""                                   (""))
+    ("-arg \"'\""                                    (""))
```
RalfJung: This (`"'"`) is an error now with my latest PR.

It may also be worth having a test for `"-arg 'val -arg val'"`, which should also error (`'` is only special *inside* `"` so this gets interpreted as `-arg`, `'val`, `-arg`, `val'`, which then errors)

